PR #31, opened 2015-07-07
Update setup.py

lowks:
Adding `description` metadata to setup, so that can be picked up by `pypi` and will show description instead of `UNKNOWN`


aptiko:
This is almost spam, judging from your github contribution history.

lowks:
Well, it's okay. Just take note that adding a `description` field will be automatically be picked up by `pypi` and be shown as a description for your project instead of listing it as `UNKNOWN`. I just saw that I thought I would help you with it. Thanks.

PR #34, opened 2015-08-17
Bootstrap3 openmeteo templates

kickapoo:


aptiko:
This is a long piece of work, so I will need to comment on each commit separately.

Let's start with the first one, 48d63e4, "Remove jquery-ui header links".

1. What are "header links"? They aren't in the header (the javascript is at the end of the html file), and I'm not certain they are called "links". You could just say "Remove jquery-ui".
2. Why are you removing `jquery-ui`? Is it not being used and has been left over by accident? Or is it your plan to replace/remove the functionality in subsequent commits? Please state so in the commit message, such as "Remove unused jquery-ui", or a long commit message with a summary of "Remove jquery-ui" and a description of "jquery-ui is only being used for xyz, and this is going to be removed in subsequent commits".
3. You seem to be removing more stuff than `jquery-ui`; namely, `tablesorter`, `placeholder`, and base.css.
4. You are also upgrading `jquery` to 1.11.3 from 1.10.2. Is this related to the removal of `jquery-ui`?

aptiko:
On 2fee39b9d, "Bootstrap 'sample-base.html'":

1. It's `base-sample.html`, not `sample-base.html`.
2. `base-sample.html` already uses bootstrap, so "Bootstrap base-sample.html" is inaccurate. "Overhaul" might be better instead of "Bootstrap". If you think that the existing bootstrap sucks (which is likely, as it was my first bootstrap) and it's fixing rather than overhaul, you could say "Fix bootstrap for base-sample.html".
3. I think `<div class="container-fluid">` should be indented, and anyway its closing tag seems to be indented differently than its opening tag, so there must be an indenting error somewhere.

aptiko:
On 4dd786c about registration templates:

1. Some templates in there have to do with emails, and nothing to do with bootstrap.
2. Why do we need to override the default registration templates? The default seem fine.

kickapoo:
I will try to answer some of your questions.
1. About 'Header links': I will rename it commit  message and will add the 'remove  jquery-ui js'. Header links was referring that are included to  inside html header tag.
2. 'jquery-ui' and all the other .css was messing bootstrap classes.
3. Update of jquery was only for personal reasons as I like to work with the latests versions.
4. Yes, email templates don't have anything to do with bootstrap but are with default registration templates i use.  Override of registration templates give us total control within the registration process. Also, 'reset password' want lead to django-admin style templates.

aptiko:
I finished my reviewing.

kickapoo:
Thank you for your review. I am working on them. 
A question based on you global comment about indentation as I know you are a Vim user. Do you have any plug in or you manually change space indentation in .html files? Is there any trick you are using ? 
I know, from your previous project that we use 2 spaces for .html files.

kickapoo:
Please review the updated  ~HEAD before I rebase branch history with squashed commits.
Also check on `station_edit.html` the restructured form. I am planning to use it to the other multi-fields forms.

kickapoo:
A force update made to remove remaining whitespaces and tabs.

aptiko:
I think I had written a huge number of comments which github ate in the forced update. That's bad. In the future I'll be writing the comments here even if it's more inconvenient.

### 1.  enhydris.css

I recall I had asked what these margins and paddings do in `body` and `navbar`. I can't see any difference without them, regardless the browser width. In which exactly situation do they make a difference?

### 2. base-sample.html

Unrelated whitespace change in line 7.

Lines 26-28: Are these empty spans doing anything?

Line 172: Is there any particular reason why you changed `<footer>` to `<div class="footer">`?

Line 183: Unrelated jquery version change.

### 3. gentityaltcode_edit.html and other gentityXXXX_edit.html files, and overseer_edit.html

Lines 10-11, 17-19 as well as the one that used to precede 8: Unrelated whitespace changes.

### 4. instrument_detail.html

Lines 14-17, 103: Unrelated changes.

Note: you can make these changes of course, and some of them (such as adding the ';' in line 23) may be important. However, they must be in separate commits, since they have nothing to do with the conversion to bootstrap.

Lines 42-91 and 106-143 also contain a large number of whitespace changes, but if it is too much work to fix, don't bother.

### 5. instrument_edit.html

Lines 8-12 and 21-23: Unrelated whitespace changes.

### 6. model_add.html

Damn, I'd like to see the comments that GitHub deleted. I wonder whether I wrote anything about model_add.html. Anyway, there are unrelated whitespace changes in line 32.

### 7. django-registration templates

Please don't add these templates, keep using the defaults offered by `django-registration-redux`.

### 8. station_detail.html

Explain in a `{% comment %}` why you have the `<br><br><br><br>`.

### 9. station_edit.html

Lines 15-17: Unrelated whitespace changes.

### 10. station_list_common.html

Unrelated changes.

### 11. success.html

Line 18: Unrelated change.

### 12. timeseries_detail.html

In the `extrajs` block there are many are unrelated changes which are difficult to enumerate here.

### 13. timeseries_edit.html

Lines 13-17: Unrelated

### 14. settings/base.py

None of these changes is needed. I recall I had explained that, but my comments have been deleted. If you find any way to access them, maybe some of them are still useful.

kickapoo:
@aptiko , based on your previous comment

About the global problem(3. & 4. & 5. 9. & 10. & 11. & 12 & 13) with unrelated whitespace/indentations. 
a) If I add in commit message  or in commits small description " .. & Fix indentations/whitespaces" , will you consider it as valid?
b) If the answer for a) is negative,  you want a separate commit  before or after bootstrap ? (This case demands to re-write from scratch the pull request)

My comments following your issue order:
1. enhydris.css: You want see any difference but is a common practice to space out content a bit when you use fixed navbar. You want me to delete them?
2.  base-sample.html
     Lines 26-28: Are these empty spans doing anything? When you are in mobile view, you see a button with three small lines. These spans are making this.
    Line 172: Is there any particular reason why you changed 'footer to div class="footer"'? No, i like to keep things clean in divs. So, i could write something like 'footer class="footer"' but I dont like it.
  Line 183: Unrelated jquery version change.: You want me to change it the previous version or just make single line commit "Update JQuery library" ?
6.  model_add: In my notes i don't have anything about this view.
7. with 14. Registration templates:  Please see here: http://django-registration-redux.readthedocs.org/en/latest/quickstart.html#templates. Sure I could delete and use django templates but those wont be bootstraped. For example, when user login, will not see bootstrap form but a classic form. The changes in settings are necessary in order to overide the django registration templates and use the custom. So, i am waiting for your final call, delete them ? yes or no?

aptiko:
### Unrelated changes

Move the unrelated changes (mostly whitespace, that is) to a different commit. It doesn't matter if it is before or after; besides, it will probably be easy to reorder, but it doesn't really matter. You don't need to redo the commit from scratch, you can [split it in two](http://stackoverflow.com/questions/1440050/how-to-split-last-commit-into-two-in-git). I think it may be easier to select the unrelated changes first (with `git add -p`) rather than the bootstrap changes, but this is up to you.

Hint: Create another branch to avoid losing your work:

```bash
git checkout bootstrap3
git branch oldbootstrap3  # Mark what we've done so far so that we don't lose it
git reset @~  # Reset, rebase, and do all kinds of crazy things on the bootstrap3 branch.
              # Our previous work is safe in oldbootstrap3.
```

### enhydris.css

What you say is not clear. I won't see any difference, but it's common to space out? Either I will see no difference, or I will see some spacing out. If there is no difference, of course you should remove; if there is some spacing out, I want to know exactly under which circumstances (which browser, which width, which page) this spacing is visible (compared to when these directives are missing), because I can't find it.

### Footer

In HTML4 we used `<div class="footer">` or `<div id="footer">`. In HTML5 it is best to use `<footer>`. http://www.w3schools.com/html/html5_semantic_elements.asp explains more. There is no reason to use the superfluous `<footer class="footer">`; if the css is `footer {...}`, it applies to the `<footer>` section.

The only reason we might need to use `<div class="footer">` is if bootstrap somehow requires it; but I don't think this is the case (but I could be mistaken).

### JQuery version

Change the jquery version in a separate commit.

### Registration templates

The default registration templates extend `base.html`, so if `base.html` uses bootstrap, so will the registration templates. However, if the registration templates also need to use other bootstrap functionality, and you think it's necessary, then keep your changes.

The changes in settings are not good, however. Some other changes are required, which I have made in https://github.com/openmeteo/enhydris/pull/38. If you approve of that pull request, tell me, so that I can merge it into master. If, in the meantime, you want to continue working, you can rebase your work on that commit, like this:

```bash
git remote add aptiko git@github.com:aptiko/enhydris.git
git fetch aptiko
git checkout templatesettings
git checkout bootstrap3
git rebase templatesettings
```
Back to the registration templates: since the templates setting uses `APP_DIRS = True`, the only change you need to do is to move `registration` to the end of `INSTALLED_APPS`, so that it is after `enhydris.hcore`.

kickapoo:
Please check once more this pull request.

aptiko:
This is almost correct. It still has minor errors (some indentation fixes during bootstrap, some wrong wording in comments), but I think that further attempting to refine such a huge piece of work would be unproductive. Therefore I will soon merge it as is.

aptiko:
I made some minor changes (a reorder, a squashing, and a little commit message rewording) before merging and GitHub got confused. In fact all changes were merged. Closing.

PR #36, opened 2015-08-29
Coordinates ui

kickapoo:


aptiko:
1. Latitude should come before longitude, and ordinate before abscissa.
2. The simple view, besides latitude and longitude, should also have altitude.
3. The "Latitude"/"Longitude" label should not have the " (WGS84)" suffix.
4. In the advanced view, the "Latitude" label should change to "Ordinate", and "Longitude" should change to "Abscissa".
5. In the advanced view, the order should be latitude, longitude, srid, altitude, asrid.
6. The label for srid should be "SRID", and for asrid "Altitude SRID".
7. It should only be possible to switch to the simple view if the srid is 4326 or null, AND asrid is null. New stations should also be possible to switch to the simple view. If the simple view is possible according to these rules, it should be the default.
8. If Javascript is deactivated, the advanced view should show.
9. When viewed in narrow widths, it is currently not clear what the switching button refers to, because the co-ordinates section is not visually separate.

kickapoo:
Please check this updated version

aptiko:
From my notes in the previous comment, you missed (1), (5) and (6).

The Javascript should go into `hcore.js`. However, before going on to move it, wait until I write about Javascript in the manual.

Note that in order to avoid using strings such as "Latitude" and "Ordinate" in the Javascript, which poses a translation problem, I wonder whether it would be easier to have two labels in the HTML, of which one will be invisible.

kickapoo:
Please check the update version. 
I also added JSCS and JSHint as most modern editors have plug in to do job.

aptiko:
1. When changing from simple to the advanced view, ordinate should become latitude (I think that the code is mostly correct; it's the label that is wrong; i.e. whenever the UI says "longitude" it actually means "latitude").
2. The labels for srid and asrid are wrong; they should be "SRID" and "Altitude SRID".
3. Since Enhydris is using JQuery, it would probably be better style to write `$('#myid')[0]` than to write `document.getElementById('myid')`. It is probably also better to use JQuery's `.show()` and `.hide()`. It's better to write this in the manual, i.e. that when an app uses JQuery, then its JavaScript will be using JQuery-style everywhere. What is your opinion on this?
4. Don't add JSCS and JSHint configuration files. Put these in your home directory or anywhere it suits you. (If it suits you in Enhydris, then put them in your personal `.gitignore`.)
5. This is in all languages: function names must be verb-like. Instead of `simpleView`, use `enableSimpleView` or `switchToSimpleView`. Likewise for `advancedView`. Instead of `coordinateSwitch`, use `toggleCoordinatesView`.
6. Write the functions in lambda format: instead of `function findLabels() {` use `var findLabels = function () {` (I will add this to the manual).
7. Don't pollute the `window` namespace. Use `var toggleCoordinatesView = function ()`, and have `enhydris.coordinatesUI` return it in the way suggested by the manual. Use `onclick="enhydris.coordinatesUI.toggleCoordinatesView()"` instead of `onclick="toggleCoordinatesView()".
8. Rename `initSwitch` to `initializeCoordinatesView`. Don't run it inside `enhydris.coordinatesUI` (as you have it it runs in all Enhydris pages, resulting in an error). Instead, return it in the way suggested by the manual and run it in an `onload` handler or in a `<script>` at the end, or add it to a handler (I don't know which is best; if you have an idea of what are best practices here we could write it in the manual).
9. The change in `station_edit.html` line 9 is unrelated.

aptiko:
With respect to (8), I think the best solution is to use `$(document).ready()` in a `<script>`.

kickapoo:
1. Done
2. Done
3. I hate jQuery, I really hate jQuery, I really really hate JQuery. But as I see from the manual i should probably use jQuery. 
4. Done
5. Done
6. I don't fully understand this one, I am waiting for the manual.
7. Ok, but window is an awesome global object. I was thinking that we can attach our translation variables.
8. Done
9. Done

kickapoo:
finders crossed, I think is ready for merge :)

aptiko:
This is much better, but there are still some issues.

1. The JavaScript has a large comment, and its last line is too long (does JSCS catch this?)
2. In the JavaScript there is a comment that says `// Initialize Switch`. I don't understand what this means, and I think that the JavaScript is perfectly understandable without it.
3. When the SRID is anything other than 4326, it should show the advanced view, and it should not be possible to click the "Advanced view" button. For example, go to a station, set ordinate=4331165.20, abscissa=245648.96, srid=2100 (this is the co-ordinates of Kompoti, it's the same thing as 39.09518 21.06071), and save. Then try to edit the station again. It shows latitude=4331165.20 etc. This is an error; the latitude is not 4331165.20; this is the ordinate.

aptiko:
In addition, it's `mapModule`, not `map_module`, and the comment about `mapModule` must be preceded by an empty line.

kickapoo:
1. JSCS  didn't catch anything.  I would like to keep this comment as it is.
2.  Ok, the comment will be deleted 
3.  Based on your previous comment:

      " It should only be possible to switch to the simple view if the srid is 4326 or null, AND asrid is null "
      `((sridValue === '4326' || sridValue !== null) && asridValue === '' ")` .
       So in the case: ordinate=4331165.20, abscissa=245648.96, srid=2100.
       `( false || true) && true` --> `true`

     So i need to alter the conditions, probably `sridValue !== null` is not needed as in other conversations is noted that users should use the default "4326" . 

4. Sorry, i didn't understand your comment about `map_module`.

kickapoo:
Updated

aptiko:
You have a comment somewhere:

    // String Variables for enhydris.map_module module

That comment must be preceded by an empty line; also, it's `mapModule`, not `map_module`.

Add `"maxLineLength": 79` to your .jscsrc (I also fixed that in the manual), and it will find the problem. You just need to wrap the last line of your big comment.

Otherwise, I think it's OK.

What I had said, "It should only be possible to switch to the simple view if the srid is 4326 or null, AND asrid is null", is correct, but it translates to this:

    ((sridValue === '4326' || sridValue === null) && asridValue === '' ")

This differs from what you say in the test in the middle, where you have `sridValue !== null` instead.

But since it appears to be working I guess you fixed that. Anyway, after we merge it I will probably write a Selenium test to debug the hell out of it, so if we discover any more problems we'll fix them then.

kickapoo:
Updated

aptiko:
I merged this (but changed its commit message).

PR #38, opened 2015-09-05
Modernize template-related settings

aptiko:
Django 1.8 deprecates the old way of specifying template settings; we
have changed settings to use the new way.


kickapoo:
From django documentation (https://docs.djangoproject.com/en/1.8/ref/templates/upgrading/) is seems that they suggest `'context_processors'` should be a `list` rather than a `tuple`.

Also, it seems that `custom context_processors` should be listed at the top of the list rather than the end.

aptiko:
I updated to use a list. However, the order does not matter.

PR #43, opened 2015-09-30
Remove the is_active fields (fixes #8)

### LINKED ISSUE #8: Remove the `is_active` fields
The `Station` and `Instrument` models both have an `is_active` field. Apparently the original designers of the database thought that it would be useful to make queries of, e.g., active stations, as opposed to all stations (including obsolete ones).

However, the correctness of this field depends on the procedures each organization has. Many organizations don't have a specific procedure for obsoleting a station; a station merely falls out of use (e.g. an overseer stops working and (s)he is never replaced). Therefore, it is unlikely that someone will go and enter the correct value in the `is_active` field. Even if an organization does have processes that could ensure correctness of `is_active`, they could merely specify an `end_date` to a station or instrument, and therefore `is_active` is superfluous.

Indeed, in all Hydroscope databases, the field seems to be randomly chosen, and in `openmeteo.org` it makes even less sense, since it is an open database whose users are expected to merely abandon their stations and not care about "closing" them properly.

However, before we remove the `is_active` fields we need to verify with all organizations participating in Hydroscope.

aptiko:


kickapoo:
So, let's make a scenario. 
I have organization and successfully installed 10 weather station to Enhydris in water division 14.  After one year my funding is cut and I can't operate my stations anymore. 

A another organization starting a study in  water division 14, enters Enhydris databae and tries to get some data. A search in database will reveal my 10 weather stations. If user search a bit deeper he/she probably check the last recording date and realize that my 10 weather stations are not supported. 

So, i believe that `is_active` must replaced with a rule like : 'If station isn't updated the last year(or 6 months) is should be marked as inactive' .

aptiko:
Yes, what you say could be accomplished by making it possible to search for stations that have been updated in the last X months/days/whatever. And it might even be a good idea to enable such a filter by default (and provide a way to override it). However, your scenario is YAGNI. It's better to wait for users to face the problem before attempting to solve it.

PR #45, opened 2015-10-02
Fix GentityEvent form submission (fixes #23)

### LINKED ISSUE #23: Internal server error when adding event
Run Enhydris locally, go to `http://localhost:8000/`, login, visit a station (e.g. `http://localhost:8000/stations/d/1372/`), select "Add new event"; fill-in date, type and user; leave the report empty. Hit "Submit changes".

Result:

> NoReverseMatch at /gentityevent/add/
> Reverse for 'station_detail' with arguments '()' and keyword arguments '{'object_id': '1372'}' not found. 1 pattern(s) tried: ['stations/d/(?P<pk>\\d+)/$']

kickapoo:


aptiko:
1. In `test_views.py`, "from enhydris.hcore.models import ...", `GentityEvent` must be before `GentityFile`, and `Timeseries` before `TimeZone`. A space is also needed before `Organization`.
2. There's a relatively new section in the manual, "Writing comments in the code". Please read it. The problem is that the comment in line 1105 applies only to the chunk of code up to 1112. Therefore, after 1112 there must be an empty line, and then another comment.
3. Assuming that `station = Station.objects.get(name='Komboti')`, it would be best, before submitting the form, to check `station.gentityevent_set.count()` before submitting the form, and after submission verify that it has increased by one. (If you do this, you won't need `station_id`, you'll have `station.id`, which you can also use in line 1109.

kickapoo:
Updated

aptiko:
Rebased and merged in 25ccb8c6f2. I made two or three minor changes (PEP8). Initially the rebased unit test was failing because of an error in `create_test_data()`, which I fixed in a subsequent commit.

PR #47, opened 2015-10-07
Fix "list stations visible on map" button

aptiko:


kickapoo:
Both commits should be merged to master

PR #49, opened 2015-10-09
Fix sort cases order validation (fixes #28)

### LINKED ISSUE #28: Internal server error in some sort cases
http://openmeteo.org/?sort=999.9

```
ProgrammingError: syntax error at or near ".9"
LINE 1: ...ptr_id" = "hcore_gentity"."id" ) ORDER BY "999999".9 DESC LI...
```

kickapoo:


aptiko:
You wouldn't write `self.assertEqual(aboolean, True)`; you'd write, instead `self.assertTrue(aboolean)`. You also wouldn't write `self.assertTrue(a < b)` but `self.assertLess(a, b)`.

The comments need improvement. You often write "search" instead of "sort". You don't need two unit tests, you can have only one, named `test_sort`. I think that two chunks of code for each request is too much; I suggest one; for example, the first request can be like this:

```python
# Sort with an invalid term; it should be ignored and result in sorting
# by name.
response = self.client.get(reverse('station_list') + '?sort=999.9')
i = response.content.index
self.assertTrue(i('Agios Athanasios') < i('Komboti') < i('Tharbad'))
```

In the main code, I find that the comments do not make the code any clearer. It's ok to just add the two new lines without any comments.

kickapoo:
Updated, 

As you will see, i kept two unit test as I think is that `test_sort` name is too general for the purpose of the Unit Test.

aptiko:
Rebased and merged in 583896356c.

PR #56, opened 2015-11-16
Add css class 'tabs-no-data' (fixes #55)

### LINKED ISSUE #55: Remove ugly hack for maintain a decent tab height
For instance:
https://github.com/openmeteo/enhydris/blob/master/enhydris/hcore/templates/station_detail.html#L270 

This should be removed with the addition of a custom `enhydris.css` class that will set `height` to a constant value.

kickapoo:
In the case were in station detail there are no data to some information tabs, class 'tabs-no-data'
add fix height to h4 element

### DIFF
--- a/enhydris/hcore/static/css/enhydris.css
+++ b/enhydris/hcore/static/css/enhydris.css
@@ -4,3 +4,8 @@
     height: 300px;
     position:relative;
 }
+
+h4.tab-no-data {
+  white-space: normal;
+  min-height:25em;
+}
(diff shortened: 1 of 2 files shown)


aptiko:
It looks ok, except for two details:

1. The comment "Empty Tabs in station_detail fixed height" has some errors, but I think it's not needed anyway, you could delete it.
2. I don't like the commit message. How about "Remove ugly hack for minimum tab height"

kickapoo:
How about now?

PR #57, opened 2015-11-17
Unit testing overhaul with model mommy

aptiko:


kickapoo:
Is this final, for reviewing? 

From that I see, I believe you over complicate things here. 
For instance, let's assume that a TestCase needs 3 different Stations with some differences in their attributes ( by attributes, I mean model.fields).

```python 
from enhydris.hcore.models import Station 
from mommy_model import mommy 

class WhateverTestCase(TestCase):
    def setUp(self):
        self.station1 = mommy.make(Station, name="Whatever")
        self.station2 = mommy.make(Station, name="Whatever", copyright_holder="Poor")
        self.station3 = mommy.make(Station, name="Whatever", copyright_holder="Rich") 
```

I believe the file `mommy_recipes.py` is not needed in such extend manner.  We have to gain from `mommy` that creates  a model instance when ever is needed and make more readable code than `create_test_data()`. Also, you should take advatage that `related objects attributes are also reachable`,
 for instance (I don't remember exactly water_division model fields)  
```python
self.station1 = mommy.make(Station, name="Whatever', water_division__name="Whatever")
```

aptiko:
I made changes similar in spirit to what you suggested about `PoliticalDivision` (but they go even further).

aptiko:
Test 1702

PR #59, opened 2016-03-31
Fix unbalanced html.div tag (fixes #58)

### LINKED ISSUE #58: Unbalanced divs
In `base-sample.html` I count 18 `<div>` and 17 `</div>`. I don't know where the error is.

kickapoo:


kickapoo:
Searching on how to validate html with django, a first suggestion is
django-html-validator (https://github.com/peterbe/django-html-validator) but this not a valid option. Even the creator has a blog post that state the problems of his solutions , http://www.peterbe.com/plog/django-html-validator.

Another solution might be a custom middleware that throughs 500 error if the page is not valid, http://bstpierre.org/Projects/HtmlValidatorMiddleware/index.html . This could easy implemented to a single UnitTest with a for loop to all enhydris views.

kickapoo:
Even that `django-html-validator` seems not to fulfil our third party library guidelines, we can consider it as valid choice due that our suggestions for custom solution share the same logic. 

@aptiko  what is your opinion? Should I proceed  with a Unit Test from `django-html-validator` and complete this pull request ?

aptiko:
I think that `django-html-validator` is worth an experiment. Note that I think you don't need to create any unit test; you just need to replace a client with a `ValidatingClient` in one of the existing tests.

kickapoo:
Closing this one as the commit is accepted
https://github.com/openmeteo/enhydris/commit/e305727ca1d4e63e3dee6723cdab462b80b21d38

Please create an issue for Unit Test that is needed.

PR #61, opened 2016-05-06
Spatialite compatibility

aptiko:
@kickapoo, I think you will love this. You can now develop Enhydris with spatialite.

If you want to review this pull request, you will find it easier to check each commit separately. (You might also find it easier to start with the last commit, the documentation changes, so that you get an idea of what has been done.)


aptiko:
I have a few comments:

1. In the documentation, there is some confusion concerning Dickinson version.
2. `from django.utils import timezone as utc`; this may work now, but it could create confusion in the future.
3. Some code-style changes in bff7d6d might not be needed.

PR #84, opened 2018-04-12
Fix ?format=csv station list view (fixes #79)

### LINKED ISSUE #79: TypeError: Transform only accepts CoordTransform, SpatialReference, string, and integer objects
cfe6044 fixed the CSV view for most cases, but appending "?format=csv" to the url produces an error when in the list of stations there is a station that has a point assigned but in which srid is null. Such a case should not exist (it may be an error elsewhere), but it does exist, so we must handle this (assuming a default srid=4326).

kickapoo:


kickapoo:
@aptiko Please review

aptiko:
Merged.

PR #86, opened 2018-10-03
Refactor

aptiko:


aptiko:
This starts looking good. I'm taking note of some remaining work so that I don't forget it:

- [ ] Add timeseries bottom
- [ ] See whether any old deactivated unit tests are needed
- [ ] Remove all remaining obsolete files
- [ ] Add option to download timeseries data in hts format
- [ ] Make warnings go away
- [ ] Upgrade Django
- [ ] Refactor documentation (and fix the release notes)
- [ ] Take care of the root URL (in development and production)

aptiko:
@kickapoo This is ready for review. Some hints:

1) Don't look at the individual commits. I'm going to reorder and squash them anyway. There are so many changes that Enhydris is almost a brand new program. Just look at the final result.
2) You might want to start by looking at the API docs, http://users.itia.ntua.gr/anthony/tmp/enhydris/dev/webservice-api.html, and then proceed to take a look at the code.
3) There are some things that I haven't fixed yet, such as fixing the release notes. I will continue while you are reviewing this. It doesn't affect your reviewing.
4) Since this is very big, we'll consider accepting it as is and adding any comments of yours as separate issues. But you needn't bother about such details at this stage.
5) If it's hard to review, make a short review. Just take a look. If you want, we might talk about it on a call, but after you take a short look.

aptiko:
Comments by @kickapoo:

1. In the docs, the examples should show the json returned.
2. Why have `urls.py` in the root directory?
3. rules should probably be moved from the bottom of `models.py` to their own file.
4. In `admin.py`, there's too much repetition of the `[f.name for f in models.Organization._meta.fields]` and similar.
5. In `csv.py` there are some `try...finally` which can probably can be better written with `with`.
6. In `csv.py`, instead of creating temporary files, could this be done in memory or so?
7. Is `initial_data` really being used?
8. There are several templates, such as `base-sample.html` which are probably not being used.

aptiko:
Note to myself:

1. Reorder the migrations so that 0005 becomes 0002, and reorder commits because the initial data migration must go right after the current master branch.
2. Fix a typo in the installation instructions where instead of "2.0" it says "1.1". This must be fixed in the 2.0 branch.
3. Write release notes and update installation instructions.

PR #87, opened 2018-11-06
Enhydris frontend initial VueJs structure

kickapoo:
Initial VueJs application structure for Enhydris frontend.
Related techs

- [yarn - package manager](https://yarnpkg.com/lang/en/)
- [vuejs cli - project creator)](https://cli.vuejs.org/guide/)


aptiko:
Instead of hello world, let's create a station list as the root page; it should be getting the data from `/api/Station`. The page doesn't need to be complete and polished; at this stage we're interested in getting the file structure and all mechanisms. So it must be unit tested. `/api` will be served by Django, `/static` will be static files, and all other URLs will be served by `vue`.

In addition, please only put JS/vue related files in `.gitignore` (no MacOSX stuff and no editor stuff).

Also, please decide on the JS standards we will be using and create an appropriate .`eslintrc` file.

PR #108, opened 2019-01-28
Improve page title (fixes #101)

### LINKED ISSUE #101: Improve page title
The list of stations has a title of "enhydris". It should be "Stations — Enhydris". The "Enhydris" part will be the same for all pages and must be configurable per installation. The delimiter between the page part (i.e. "Stations") and the site part (i.e. "Enhydris") is space/em-dash/space. The em-dash is U+2014.

(But if nuxt happens to do that automatically, it's fine if it uses another character instead).

kickapoo:


kickapoo:
@aptiko Yeap, I have updated : https://github.com/openmeteo/enhydris/pull/108/commits/b87a2522982e06aad3f1d036618bfc277fb4f25e#diff-8fb20d8c363337bb907e427fea5086b4R3

PR #127, opened 2019-03-18
Remove deletion of user profile from migrations (fixes #122)

### LINKED ISSUE #122: Remove deletion of UserProfile from migrations
Child of aptiko/openhi#61.

Enhydris 2 uses the user profile table, but Enhydris 3 does not. For a while we will be running Enhydris 2 and Enhydris 3 from the same database, so we don't want to delete the table yet in order to allow Enhydris 2 to continue functioning without change in this respect.

aptiko:


aptiko:
Totally wrong pull request, closing.

PR #134, opened 2019-03-26
Implement ENHYDRIS_USERS_CAN_ADD_CONTENT=False

### LINKED ISSUE #133: Permissions system does not take into account the case of ENHYDRIS_USERS_CAN_ADD_CONTENT=False
Child of aptiko/openhi#13.
Blocks #131.

aptiko:


enhydris/rules.py
```diff
@@ -1,53 +1,113 @@
+from django.conf import settings
+from django.contrib.auth.backends import ModelBackend
+
 import rules
 
 
 @rules.predicate
 def is_station_creator(user, station):
-    return station.creator == user
+    return (station is not None) and station.creator == user
```
kickapoo: I dont understand the case of station None. You have a UnitTest checking this insanity?

aptiko:
@kickapoo I made some changes which made the "is None" check no longer needed, and I also documented how it works. It's all in a new commit (which after approval I will squash into the previous one).

PR #143, opened 2019-04-09
Add timeseries inline to StationAdmin (fixes #139)

### LINKED ISSUE #139: Add time series to StationAdmin as inline

aptiko:


aptiko:
@kickapoo This sits on top of #140, so look at that first.

PR #148, opened 2019-04-10
feat(components): Add Language Picker (fixes #112)

### LINKED ISSUE #112: Improve language switcher
Child of aptiko/openhi#7.

The language switcher should be made a Vue component.

It shouldn't have the Greek hardwired; instead, it should get the available languages from configuration.

It should be made to look better. My opinion is that something like `EN EL FR DE`, where the current language is bolded and unclickable, would be adequate.

kickapoo:
Screenshot 2019-04-10 14 06 16


frontend/tests/components/layout/NavBar.spec.js
```diff
@@ -0,0 +1,26 @@
...
+        $i18n: { locale: "en" }
+      }
+    });
+  });
+
+  it("has default language en", () => {
+    expect(cmp.vm.currentLanguage).toEqual("en");
+  });
+
+  it("render the $t('home')", () => {
```
aptiko: Should be `"renders $t('home')"`.

frontend/tests/components/layout/LangPicker.spec.js
```diff
@@ -0,0 +1,23 @@
...
+        $i18n: { locale: "en" },
+        switchLocalePath: msg => msg
+      },
+      stubs: {
+        NuxtLink: RouterLinkStub
+      }
+    });
+  });
+
+  it("renders correctly with a snapshot", () => {
```
aptiko: Unless "renders correctly with a snapshot" is standard practice elsewhere, let's just use "renders correctly".

frontend/tests/components/layout/NavBar.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    expect(cmp.find("#home").element.text).toContain("home");
+  });
+
+  it("renders correctly with a snapshot", () => {
```
aptiko: See another review comment elsewhere.

frontend/tests/components/layout/LangPicker.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    // console.log(cmp.element);
+    expect(cmp.element).toMatchSnapshot();
+  });
+});
```
aptiko: While I don't understand exactly how `LangPicker` works, I understand that you can click on it and change language. We should test that such an action works.

aptiko:
@kickapoo Please squash the commits. If you have permissions, merge into master and push to official.

PR #162, opened 2019-04-19
Add jest script commands

kickapoo:
2 scripts command for unit tests

`yarn tests:unit` and `yarn tests:watch`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ matrix:
        - yarn install
       script:
        - yarn lint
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -64,7 +64,7 @@ matrix:
        - yarn install
       script:
        - yarn lint
-       - yarn run jest
+       - yarn run test
```
aptiko: Shouldn't this be `yarn test`?

PR #169, opened 2019-05-03
Capitalize env variable (fixes #165)

### LINKED ISSUE #165: adminLink should be ADMIN_LINK
So far the configuration variables in `.env` have been in all caps; but `adminLink` is different.

kickapoo:


frontend/.env.sample
```diff
@@ -2,3 +2,4 @@ NPM_CONFIG_PRODUCTION=false
 HOST=0.0.0.0
 NODE_ENV=development
 BASE_API_URL='http://stage.openmeteo.org/api/'
+ADMIN_LINK = ''
```
aptiko: Is `.env` a shell file? Then the `=` sign should not have spaces before or after it.

PR #311, opened 2020-06-02
Update default.html template with Bootstrap v4

ad0v0:
Seems to be working fine. Please review when you have time for it. Thank you! :)


enhydris/templates/enhydris/base/navbar-default.html
```diff
@@ -1,29 +1,24 @@
...
+  </button>
+
+  <div class="navbar-collapse collapse" id="navbarSupportedContent">
+    {% block search %}
+      {% include "enhydris/base/search.html" %}
+    {% endblock %}
+    <ul class="navbar-nav ml-auto">
+      {% if user.is_authenticated %}
+      <li class="nav-item"><a class="nav-link" href="{% url 'admin:index' %}">{% trans "Dashboard" %}</a></li>
+      <li class="nav-item"><a class="nav-link" href="{% url 'admin:logout' %}?next={{request.path}}">{% blocktrans with username=user.username %}Logout {{username }}{% endblocktrans %}</a></li>
```
aptiko, line 18: Could you indent these? Likewise the ones in the `{% else %}`.

enhydris/templates/enhydris/base/navbar-default.html
```diff
@@ -1,29 +1,24 @@
...
+      {% else %}
+      <li class="nav-item"><a class="nav-link" href="{% url 'admin:login' %}?next={{request.path}}">{% trans "Login" %}</a></li>
+      {% endif %}
+    </ul>
+  </div> <!-- /.navbar-collapse -->
```
aptiko, line 23: I think we don't need the comment here about which `div` we're closing, because there's no other `div` around.

I'd like to take the opportunity to ask you something, however. You may have noticed that I keep the HTML quite small, making heavy use of `{% include %}`. I started this practice some time ago (I was admittedly influenced by R.C. Martin's "Clean Code", but he's talking about programming code, not markup languages). Enhydris's templates were in a mess, and the indentation was inconsistent. I made a change somewhere, and something wasn't working because there was a missing or an extra `div`. I was unable to find it. So I sat down and overhauled the templates and made them like they are today. The guideline I followed is that if it doesn't look extremely simple, if you need to put a comment to mark which `div` you're closing, or if the closing `div` is not very easy to see and only a few lines away, then you must take the content of that `div` and put it in another file. (Actually I'd prefer it if it was in the same file, in a kind of "function", like we do when programming, but Django doesn't offer that possibility.)

Abdullah though found this practice strange. He said that according to his experience template files are large. But he's not a HTML developer either. I'm not quite certain what rule to follow. What's your opinion?

enhydris/templates/enhydris/base/search-default.html
```diff
@@ -1,25 +1,32 @@
...
-    <li>
-      <a href="#searchtips" data-toggle="modal">
-        <i class="glyphicon glyphicon-eye-open"></i> {% trans "Search tips" %}
+  <ul class="navbar-nav">
+    <li class="nav-item">
+      <a class="nav-link" href="#searchtips" data-toggle="modal">
+        <svg class="bi bi-eye" width="1em" height="1em" viewBox="0 0 16 16" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
+          <path fill-rule="evenodd" d="M16 8s-3-5.5-8-5.5S0 8 0 8s3 5.5 8 5.5S16 8 16 8zM1.173 8a13.134 13.134 0 0 0 1.66 2.043C4.12 11.332 5.88 12.5 8 12.5c2.12 0 3.879-1.168 5.168-2.457A13.134 13.134 0 0 0 14.828 8a13.133 13.133 0 0 0-1.66-2.043C11.879 4.668 10.119 3.5 8 3.5c-2.12 0-3.879 1.168-5.168 2.457A13.133 13.133 0 0 0 1.172 8z"/>
+          <path fill-rule="evenodd" d="M8 5.5a2.5 2.5 0 1 0 0 5 2.5 2.5 0 0 0 0-5zM4.5 8a3.5 3.5 0 1 1 7 0 3.5 3.5 0 0 1-7 0z"/>
+        </svg> {% trans "Search tips" %}
```
aptiko, line 29: This svg is there in order to give you the eyelike image? I'm not an expert of course, but it seems strange to me to have all this HTML code just to create an icon. Is it possible to put this in the CSS file, or in another file, which will then be referenced in the HTML with `img` or something?

ad0v0:
Concerning this [comment](https://github.com/openmeteo/enhydris/pull/311#pullrequestreview-422733625): for now I've only upgraded the default template , page navbar and search-tips modal  with bootstrap 4. Will move to page content tomorrow.

enhydris/templates/enhydris/base/search-default.html
ad0v0: This is an inline SVG  that allows us to change the color using CSS.
- If I remove it from HTML then I'll have to create two images with different colors in order to make color change effect when hovering over the item
- But, I can create an SVG sprite with this tool https://github.com/jonathantneal/svg4everybody and there's going to be less code in HTML . But we'll have to load onemore external resource

Which solution  do you preffer?

enhydris/templates/enhydris/base/navbar-default.html
ad0v0: Well, this was just my habit to comment the closing tag for all divs, as they doesn't contain any specifi meaning from a semantiс point of view. And with the comment it's easier for me to track what exactly I have here.

As for the other part of the comment: I understand what you mean, but sometimes Bootstrap has quite a deep nesting of divs withing divs and ussually it is done so for styling purposes. Thought, sometimes there's a semantic meaning as well. And breaking the code into different files would probably lead us to creating more and more HTML files. I'm not sure which is worst in this case: having dozens of lines inside one file or having dozens of files inside one folder. For example, if we take a modal component and break its content into header/body/footer parts (not to mention that it also contains dialog and content divs as well http://prntscr.com/ssj33a), instead of one file, we're going to have 4 files for it. And there's always a possibility to add new modals to the page/site. Hope this makes sense. As an option , I can replace some divs with other semantic tags, such as section/header/footer. This might improve the readibility of the code, but I'd probably keep everything inside one file

ad0v0:
⚠️ Can't find footer with links (http://prntscr.com/stujk0) in templates folder and for some reasons it won't render on localhost either (http://prntscr.com/stujz8). Openmeteo: https://openmeteo.org/

ad0v0:
🔲 Find missing footer
🔲 Resolve discussions in the existing pull request
🔲 Fix data rendering (data in tabs and card on the same row with map on the station details page)
🔲 Upgrade templates in timeseries_details to Bootstrap v4

enhydris/templates/enhydris/base/search-default.html
ad0v0: @aptiko Please review the message above when you get a chance!

enhydris/templates/enhydris/base/default.html
```diff
@@ -43,13 +43,13 @@
     <script src="{% static 'js/jquery-3.5.1.min.js' %}"></script>
     <script src="{% static 'js/popper.min.js' %}"></script>
     <script src="{% static 'js/bootstrap.min.js' %}"></script>
+    <script src="{% static 'js/read-more-js.js' %}"></script>
```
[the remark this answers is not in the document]
ad0v0: Actualy, it wasn't int he orignal version. I've added this script only last week :) But I've got your point and will fix it

enhydris/static/js/enhydris.js
```diff
@@ -147,3 +147,8 @@ enhydris.coordinatesUI = (function namespace() {
         toggleCoordinatesView: toggleCoordinatesView
     };
 }());
+
+$readMoreJS.init({
```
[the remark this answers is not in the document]
ad0v0: @AbdullahKady And I need your opinion on separating JS code from HTML wherever it is possible

enhydris/templates/enhydris/station_list/table.html
```diff
@@ -3,7 +3,7 @@
 {% load humanize %}
 
 <div class="table-responsive">
-  <table class="table table-hover">
+  <table class="table table-hover home-table">
```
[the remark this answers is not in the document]
ad0v0: Makes sense, thanks!

enhydris/templates/enhydris/station_detail/tabs/gentity_files.html
```diff
@@ -23,7 +23,7 @@ <h4 class="my-3">{% trans 'Station (Gentity) Files' %}</h4>
             <td>{{ file.date|date:"Y/m/d" }}</td>
             <td>{{ file.content }}</td>
             <td>{{ file.descr }}</td>
-            <td>{{ file.remarks }}</td>
+            <td class="show-more"><p>{{ file.remarks }}</p></td>
```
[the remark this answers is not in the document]
ad0v0: The outup with `linebreaks` looks odd wheter I remove parapgraph tag or leave as is: http://prntscr.com/t01oax and http://prntscr.com/t01ol9 .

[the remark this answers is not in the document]
aptiko: Maybe it's a bug in `readmore`, or maybe it's not designed for this, or maybe you aren't using it correctly. What's certain is that the remarks/report must have paragraphs—it's fine as it would show in the screenshots if there weren't all these "read more" links.

Maybe a `<div>` would help?

    <td><div class="show-more"><p>{{ file.remarks|linebreaks }}</p></div></td>

Not certain whether it needs the `<p>` tags.

ad0v0:
@aptiko Is it fine to have the tables like this http://prntscr.com/t01tsd (gentity events and files) on small devices? Or should I modify the look so that all the the content for each item is visible?  Like it's done for the timeseries tab: http://prntscr.com/t01wlh

aptiko:
> @aptiko Is it fine to have the tables like this http://prntscr.com/t01tsd (gentity events and files) on small devices? Or should I modify the look so that all the the content for each item is visible? Like it's done for the timeseries tab: http://prntscr.com/t01wlh

It's obviously not fine, but we won't fix that yet. Let's stick to upgrading to bootstrap 4 in this issue, and then we'll start improving things.

enhydris/templates/enhydris/base/search-default.html
ad0v0: @aptiko I'm going to leave it as is for now. Let me know if this is not fine!

enhydris/templates/enhydris/base/navbar-default.html
ad0v0: @aptiko I've checked the [Google Style Guide](https://google.github.io/styleguide/htmlcssguide.html) and the `end = <!-- /.class-name -->` comments don't violate the rules, so I'm planning to leave them as is for now. Same here as well - let me knwo if this disturbs you! And as for code otimization, I can do that after we're done with Bootstrap migration

ad0v0:
@aptiko Linebreaks: there's an original code with linebreaks http://prntscr.com/t034jf and the output for this block http://prntscr.com/t035vh is the same I had in the [tables ](https://github.com/openmeteo/enhydris/pull/311#issuecomment-644083928)

ad0v0:
@aptiko http://prntscr.com/t061as There's a placemarks.kml file in station_list folder. Should I do anything with it?

enhydris/static/js/read-more-js.js
```diff
@@ -0,0 +1,128 @@
+/**
```
AbdullahKady, line 1: I have a small concert regarding this:
- I figured this is not your code (basically like importing jQuery), so I searched for ReadMoreJS and the first thing that popped up was [this](https://github.com/jedfoster/Readmore.js/), which seems much more popular, so any reason for not using it?
- I think [this](https://github.com/georapbox/ReadMore.js/blob/master/src/readMoreJS.js) is the one actually being used? if so, can you please credit the link since the author's link is not working, also please use the minifed version if we are not editing it.

enhydris/static/js/enhydris.js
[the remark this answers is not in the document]
AbdullahKady: @aptiko @ad0v0 I'm actually really confused with that.
Using a dollar sign for variable naming is an old convention that some people used to denote that the variable holds a jQuery object like: `let $name = $('.name')`
Now I'm not sure why the author did this? I looked through the code and they are not even using jQuery in their module.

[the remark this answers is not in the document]
AbdullahKady: Also as for separating JS from HTML, I definitely second that! 
However since we are not using a bundler/transpiler, I think adding minimal JS inside the HTML will be advised in some cases, for instance the following snippet you added seems better inside the HTML rather than an external file
```javascript
    $readMoreJS.init({
      target: '.show-more p',
      numOfWords: 25
    });
```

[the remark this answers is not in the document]
ad0v0: @AbdullahKady  Got it! 

`$` was used for plugin customization in Demo, so I used it in the project as well. Should I leave it as is for now or fix it? I've used this plugin: https://www.cssscript.com/demo/read-more-read-less-functionality-in-pure-javascript-readmore-js/. Let me know if it's okay to keeps it. But if you have any other ideas or solutions on how this read more/less functionality can be implemented, let me know! I'm open to it

As for separating JS from HTML, do you wish me to do it or do you want do it yourself?

enhydris/static/js/read-more-js.js
ad0v0: Okay, it seems [your solution](https://github.com/jedfoster/Readmore.js/) is better. I'll use that instead. Thanks! And I'm going to use minified version as well

ad0v0:
Concerning your comment [here](https://github.com/openmeteo/enhydris/pull/311#pullrequestreview-431413191):

enhydris/static/js/enhydris.js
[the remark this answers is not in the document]
AbdullahKady: I believe they were already separated for the most part, no?

Also @aptiko I'm not so proficient with Django's methodology when it comes to templates and static files in general since I have been using it so far for APIs primarily. Is there anyway we could have 'relative' static files? so let's say we have the following directory structure:
```
root/
├── chart/
│   ├── index.html
│   ├── index.js
│   └── index.css
└── some-other-page/
    ├── index.html
    └── index.js
```
and in the `chart/index.html` we can directly reference JS as `./index.js`?

AbdullahKady:
> Concerning your comment [here](https://github.com/openmeteo/enhydris/pull/311#pullrequestreview-431413191):
> 
> * Yes, I've only added the code with more/less functionality. All other code is from original version of the project
> * Understood about `vendor` folder
> * As for CDN, I prefer not to rely on external resources which I can't control, whenever it's possible to do. This is probably the only reason why I haven't used CDN

@ad0v0 no idea why Github doesn't support threads/replies for generic comments, it's annoying. anyway:
For the CDN, I would personally argue that using a reputable CDN such as cdnjs or google's will most likely outperform our own hosting. It's definitely more optimized, and is likely to be more reliable too.

enhydris/static/js/enhydris.js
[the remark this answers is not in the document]
ad0v0: You mean most of the code is already separated? Well, yes, but no :) Some still in a mess:
- `timeseries_detail/chartjs.html` for some reasons only contain JS code inside HTML file
- `station_detail/main-default.html` has some JS code in the middle of the file: http://prntscr.com/t1k2ut. Scripts are short, but maybe it's possible at least move the code at the end of the file?

Okay, I've overreacted 😅 There're only two files that concern me

ad0v0:
@AbdullahKady I've totally failed with implementation of [this](https://github.com/jedfoster/Readmore.js) readmore plugin 😑 Could you please take a look at it? You can test the result on http://localhost:8001/stations/1334/ (events and files tab). For some reason it won't react to customization in script after loading the plugin http://prntscr.com/t1kxqg

AbdullahKady:
> @AbdullahKady I've totally failed with implementation of [this](https://github.com/jedfoster/Readmore.js) readmore plugin  Could you please take a look at it? You can test the result on http://localhost:8001/stations/1334/ (events and files tab). For some reason it won't react to customization in script after loading the plugin http://prntscr.com/t1kxqg

Just tried it out, it seems that it works only on elements that are in the view (aka when switching to the files tab and re-executing it, it works). I had a bigger problem, which is the functionality is not as expected at all, the styling gets messed up terribly (overlapping elements).

I guess you can just use the library you had installed previously if it worked fine, however I have a suggestion, why not do it in a purely overflow fashion (a scroll pane when it reaches a certain height), I feel like that's more robust since it's inside a table, and expanding "read more" leads to uneven row lengths. What do you think from UX point of view? since I don't have a good sense for that :sweat_smile:

ad0v0:
@AbdullahKady Hi! So, I've just tried to replace the existing scripts and styles that are served from our side with CDN's styles and script. If I use a slim version of Jquery, there's a TypeError in the console: http://prntscr.com/t39zwl. If I replace a slim version with standard Jquery, there're even more errors: http://prntscr.com/t3a0gb. I've searched for a solution on the Internet but was not able to find anything - most of the people suggest to use standard Jquery which in our case leads to more errors. If you have any solutions, let me know, please! And I've replaced CDNs with and old solution when we loaded script etc on our side.

Also, in case if we're going to be able to serve code from CDN, I think we'll need to implement fallbacks for scripts and styles to mimizу risks when something goes wrong on CDN's side. What do you think? As for scripts, there're options to choose from: for example, here is what Hanselman suggests: https://www.hanselman.com/blog/CDNsFailButYourScriptsDontHaveToFallbackFromCDNToLocalJQuery.aspx. But as for styles, I wasn't able to find anything and I don't think I've ever done that in my work. Would be very grateful if you could share your point of view on this matter!

ad0v0:
> I guess you can just use the library you had installed previously if it worked fine, however I have a suggestion, why not do it in a purely overflow fashion (a scroll pane when it reaches a certain height), I feel like that's more robust since it's inside a table, and expanding "read more" leads to uneven row lengths. What do you think from UX point of view? since I don't have a good sense for that

Well, indeed this would look a little bit old-school and considering the fact that Antonis wants us to improve the website from UI side as well, to make the website look newer and fresher. So I think we should avoit such things. Not sure if I was able to find proper words for that 😑

AbdullahKady:
> @AbdullahKady Hi! So, I've just tried to replace the existing scripts and styles that are served from our side with CDN's styles and script. If I use a slim version of Jquery, there's a TypeError in the console: http://prntscr.com/t39zwl. If I replace a slim version with standard Jquery, there're even more errors: http://prntscr.com/t3a0gb. I've searched for a solution on the Internet but was not able to find anything - most of the people suggest to use standard Jquery which in our case leads to more errors. If you have any solutions, let me know, please! And I've replaced CDNs with and old solution when we loaded script etc on our side.
> 
> Also, in case if we're going to be able to serve code from CDN, I think we'll need to implement fallbacks for scripts and styles to mimizу risks when something goes wrong on CDN's side. What do you think? As for scripts, there're options to choose from: for example, here is what Hanselman suggests: https://www.hanselman.com/blog/CDNsFailButYourScriptsDontHaveToFallbackFromCDNToLocalJQuery.aspx. But as for styles, I wasn't able to find anything and I don't think I've ever done that in my work. Would be very grateful if you could share your point of view on this matter!

Hey, sorry for the late reply, had some family problems in the past couple of days.

So for CDNs, I didn't think of fallbacks because my idea was, our hosting is more likely to go down than the CDN's.
So I was suggesting it for simplicity. 
If we need to implement fallbacks (which totally make sense), I think it's better to ignore CDNs for now (maybe future task?), to avoid the hassle of both: implementing fallback mechanism(s), and fixing the errors with jQuery that you talked about (Unless of course Antonis has priority for that). Here is a much simpler -hacky- approach for [script-fallback](https://stackoverflow.com/questions/5257923/how-to-load-local-script-files-as-fallback-in-cases-where-cdn-are-blocked-unavai) that I found. As for CSS fallback, I found some clever (but following the same hack-y way of script fallbacks), you can check [1](https://stackoverflow.com/questions/50884212/fallback-to-local-bootstrap-4-css-file), and [2](https://stackoverflow.com/questions/7383163/how-to-fallback-to-local-stylesheet-not-script-if-cdn-fails)

enhydris/templates/enhydris/base/navbar-default.html
aptiko: @ad0v0 The problem is we shouldn't be keeping our own habits. Otherwise if another person comes and has a different habit, some closing divs will have a comment (even if obvious, as in this case), and some will be without comment, although it might be necessary in that case. And a third person will write comments but without the slash and the dot that you are using (what does that mean, btw?).

I'll try writing something in the SOPs and see if we can avoid a flame war :-)

enhydris/templates/enhydris/base/search-default.html
aptiko: @ad0v0 Why do we need svg4everybody? Is browser support insufficient for loading external svgs?

enhydris/static/js/enhydris.js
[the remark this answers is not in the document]
aptiko: @ad0v0 Yes, chartjs.html is a big mess and we're about to rewrite that from scratch.

@AbdullahKady I'm not certain what you mean, but I think you have that wrong. Templates are put in a different directory than static files. In production, static files are copied to a directory and served from there by the web server, without involving Django at all. Templates, however, need to be rendered by Django. I don't think you can reference static files in any relative way—you have to write `{% static 'enhydris/js/chart.js' %}` to access the static file.

If this isn't clear, [Figure 5.2 of Deploying Django](https://djangodeployment.readthedocs.io/en/latest/05-static-files.html#how-static-files-work-apache) might help.

enhydris/templates/enhydris/base/navbar-default.html
aptiko: @ad0v0 Try this: https://wiki.antonischristofides.com/en/sops/production/development/code-style#breaking-large-files

I think one key part is how much the editor helps you. For example, some editors draw vertical lines to help you identify the extent of each element. My editor (vim) doesn't do this. Let me see how the example looks with sublime... Hmm, I don't think it helps very much. I think that folding helps only a little. What I think would help is if it could somehow change the background color of the entire element you're looking at (like when you select it, for example).

So, for the time being, let's follow the "breaking large files" guideline, which in this case means remove the comment at the closing div. And let's flame each other about that guideline in Slack :-)

aptiko:
@AbdullahKady @ad0v0 In Enhydris we have been using a CDN (cdnjs.cloudflare.com specifically). So let's continue using that.

While working you come across an issue (e.g. whether to use or not use a CDN). You can check what we already do. In this case we have been using cloudflare. It follows that you should follow the established practice. If you disagree with the established practice (you're welcome to do so! I don't want our code to use ancient practices just because no-one brings up such issues), then you should bring up your disagreement in Slack or file another issue. I think that discussion whether to use a CDN or not in this issue is counter-productive (unless the established practice isn't clear from the code).

ad0v0:
@AbdullahKady Thanks for the links! At first glance, solutions are good, but you never know what pifull you're ging to face until you begin implementing those solutions 😵

enhydris/templates/enhydris/base/search-default.html
ad0v0: @aptiko I think the main issue here was to reduce the amount of code in HTML for each image and to do that I'll have to generate sprite.svg file that combines all *.svg files inside it. This is why I suggested to use svg4everybody

enhydris/templates/enhydris/base/navbar-default.html
ad0v0: @aptiko 
> And a third person will write comments but without the slash and the dot that you are using (what does that mean, btw?).

dot means class and slash means closing tag :)

ad0v0: Also, another option is to use "end" word instead of slash. For example, ''

enhydris/templates/enhydris/base/default.html
```diff
@@ -1,48 +1,65 @@
...
-      enhydris.rootUrl = '{% url "station_list" %}';
-    </script>
-    <script type="text/javascript" src="{% static "js/enhydris.js" %}"></script>
-  {% endblock %}
-  {% block extrajs %}{% endblock %}
-</body>
+
+    {% block extrajs %}{% endblock %}
+  </body>
+</html>
```
aptiko, line 65: Files should end in a newline. This should be configurable in your editor.

enhydris/templates/enhydris/base/main_with_map-default.html
```diff
@@ -8,12 +8,13 @@
 
 {% block mainjs %}
   {{ block.super }}
-  <script src="//cdnjs.cloudflare.com/ajax/libs/leaflet/1.5.1/leaflet.js"></script>
+  <!--TODO Check for the new versions of the scripts below -->
```
aptiko, line 11: No TODOs.

enhydris/templates/enhydris/base/search-default.html
aptiko: I don't care about size so much. I find it unnatural for the HTML to contain such code. If it's easy to put it in a separate svg file and load it with `` or something, let's do it this way. Otherwise we can keep it as it is. No need to add a plugin.

enhydris/templates/enhydris/timeseries_detail/chartjs.html
```diff
@@ -1,11 +1,12 @@
 {% load static %}
 
+<!--TODO check for new versions of the scripts below -->
```
aptiko, line 3: No TODOs. I also don't think there's any reason to check for new versions if the existing ones aren't giving us any headaches. flot, in particular, will be dealt with by a JS programmer when the time comes to rewrite the front-end for the charts (hopefully quite soon, that is).

AbdullahKady: @ad0v0 Also, for chartjs.html specifically, please revert the changes if it's non-functional (only semantics/structuring), since it will conflict heavily with my current open PR.

ad0v0:
> You have added a static/js/vendor directory after all. Could you use cloudflare instead?

I've added vendor directory before you've shared that magical link that solved the error :D But still, I think it would be better to leave vendor folder with  all the files. In order to use it later so we could create a fallback in case is something goes wrong with CDN. But if you don't like this idea, then I can remove the directory

enhydris/templates/404.html
```diff
@@ -1,17 +1,11 @@
...
-replaced with all sorts of unhelpful nonsense. So they would ignore
-all useful information on the server's response and put up their own
-404 page instead. I hear Google Chrome followed that incredible
-stupidity. Lucky they decided to only do it for pages less than 512
-bytes in length, which is the reason for this comment's existence.
--->
-</body>
+<!DOCTYPE html>
+<!--TODO Add proper lang value -->
+<html lang="">
```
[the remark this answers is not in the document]
ad0v0: @aptiko  I've probably missed why we can't use TODO. Could you explain it once more please? 
PS will it be okay if I leave lang attribute without value for now? Or maybe you coud help with proper value? I've tried Googling first, but Google refuced to cooperate 😑

enhydris/templates/404.html
```diff
@@ -1,17 +1,11 @@
...
+  <head>
+    <title>Object not found</title>
+  </head>
+
```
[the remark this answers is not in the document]
ad0v0: @aptiko  I think there's no issue with the length of the content on the page. What I thought was wrong is that we had a commented text that is irrelevent to the content of the website. This is why I've removed it. But... if we have any resources later I'd suggest updating 404 etc pages to look more infromative and good looking. You can see some exmaples here: https://www.interaction-design.org/literature/article/how-to-design-great-404-error-pages. By the way, this is a great resource to learn UX design if you have any desire... and time, of course :)

enhydris/templates/enhydris/base/default.html
ad0v0: Configured settings

enhydris/templates/enhydris/timeseries_detail/chartjs.html
ad0v0: @aptiko @AbdullahKady I haven't changed the version of the scripts. Only changed the link from `//url.name` to `https://url.name` as it is said in Google code guide which we should follow. Hope I'm not mixing things and correctly understood the requirements

ad0v0:
> I don't care about size so much. I find it unnatural for the HTML to contain such code. If it's easy to put it in a separate svg file and load it with  or something, let's do it this way. Otherwise we can keep it as it is. No need to add a plugin.

@aptiko Yes, I understand it may look a little bit odd when you face such huge code for only one image (btw it's actually quite small :D in SVG code context) in HTML file, but using SVG now (and sometimes this way) is a normal practise. Let's leave it as is for now. I'll see how many icons there will be in the end for this project. For now it's much easier to style an SVG image when it's inlined directly into HTML file. And one more good resource: https://svgontheweb.com/

enhydris/templates/enhydris/timeseries_detail/chartjs.html
AbdullahKady: @ad0v0 yea I know, but I'm hoping that you can revert that for this very specific file, I can add the `https://` myself in the PR. It's just an unnecessary conflict that one of us will have to deal with -depending on which PR is merged first-.

ad0v0: Done :)

enhydris/templates/enhydris/base/messages.html
```diff
@@ -1,10 +1,14 @@
-{% if messages %}
-<div class="alert alert-info alert-dismissible" role="alert">
-    <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-    <ul>
-      {% for message in messages %}
-        <li{% if message.tags %} class="{{ message.tags }}"{% endif %}>{{ message }}</li>
-      {% endfor %}
-    <ul>
-</div>
+{% if message %}
```
aptiko, line 1: messages

aptiko:
Merged in 9461257.

PR #313, opened 2020-06-10
Rewrite time series chart backend code (fixes #181)

### LINKED ISSUE #181: Rewrite time series chart backend code
The existing time series chart code is completely unreadable and needs to be rewritten. To find it, grep the code for 181.

In order to do it properly, we need to add an API endpoint: `/api/stations/XXX/timeseries/YYY/chartdata/`. This is the equivalent to the old `/timeseries/data/?object_id=YYY` url.

What the old endpoint appears to be doing is return a json object in the following format:

```
{
    "data": [
        [913307400000, "6.6", 1],
        [916097400000, "15.2", 4651],
        [918887400000, "16.8", 9301],
        ...
    ]
}
```
(Except for `data` the returned object also contains `stats`, but it appears to be unused.)

The plot of a time series consists of 201 points. (More precisely, it consists of `ENHYDRIS_TS_GRAPH_BIG_STEP_DENOMINATOR + 1` points. The setting has the default value 200, but we'll change this to a hardwired 200, because it doesn't make sense for this to be configurable.)

We take the length of the time series and divide it by 200 intervals, resulting in 201 data points including the first and the last one. A more correct way would be to take the start and end date and divide them in 200 intervals, resulting in 201 time stamps. In addition, the old code seems to be including more or less than 200 points, depending on whether there are missing records, but I can't figure out the details.

Each item in `data` consists of three items: timestamp, value, position.

The **timestamp** is in milliseconds since the epoch. The **position** is the one-based index of the time series record. In the old code, the positions rather than the timestamps appear to be equidistant, and the `timestamp` is is the timestamp of the time series record with that `position` (but it might have been more correct for the timestamps to be equidistant and for the `position` to be the index of the record with the timestamp closest to `timestamp`).

The **value** is the value to plot. I can't figure out how the old code determines that. It could be the max or the average of a group of values near the time series record at `position`.

The old endpoint accepts arguments `start_pos` and `end_pos`, which are used when zooming in.

The old JS code is smart enough to stop drawing a line in presence of null records but not smart enough to do so in the absence of records.

AbdullahKady:


enhydris/api/views.py
```diff
@@ -183,17 +184,62 @@ def bottom(self, request, pk=None, *, station_id):
...
+        result = []
+        for _ in range(number_of_samples):
+            try:
+                idx = df.index.get_loc(
+                    current_time, method="nearest", tolerance=tolerance
+                )
+                value = df.iloc[idx].value
+            except KeyError:
+                # If no matching index is found (ie exceeds tolerance)
+                # a KeyError is raised rather than a graceful null return
```
aptiko, line 223: Please write comments with the same care as code. Use "i.e." instead of "ie" and make sure the syntax is correct, e.g. "the tolerance is exceeded" instead of "exceeds tolerance" and "returning null" instead of "graceful null return".

enhydris/api/views.py
```diff
@@ -183,17 +184,62 @@ def bottom(self, request, pk=None, *, station_id):
...
+        number_of_samples = min(
+            settings.ENHYDRIS_TS_GRAPH_BIG_STEP_DENOMINATOR, len(df.index)
+        )
+        min_time = df.index.min()
+        # To accommodate the first data point
+        current_time = min_time
+        interval = (df.index.max() - df.index.min()) / (number_of_samples - 1)
+        tolerance = interval / 2
+        result = []
+        for _ in range(number_of_samples):
```
aptiko, line 215: This looks very strange. The loop is repeated a fixed number of times and each of these times the `current_time` variable changes? Can't the loop be changed to `while current_time < ...`?

BTW, I don't think it's a good idea to use `_` for an unused variable, because of a potential conflict with `gettext()`. But unused variables often indicate a possible improvement in the code.

enhydris/api/views.py
```diff
@@ -183,17 +184,62 @@ def bottom(self, request, pk=None, *, station_id):
...
+        )
 
+    def _sample_equidistant(self, df):
+        """
+        Divides dataframe/timeseries by time into "number_of_samples" equally,
+        and returns a list of dicts, where each dict contains timestamp, value
+        """
+        number_of_samples = min(
+            settings.ENHYDRIS_TS_GRAPH_BIG_STEP_DENOMINATOR, len(df.index)
+        )
```
aptiko, line 208: It's not entirely clear to me from its name and docstring what this function is doing. I understand, however, that it's probably doing something that may be hard to capture in a name. Maybe `_sample_equidistant` describes the how and not the what? Maybe something along the lines of "get values to plot" is the what?

The docstring should also be compliant to the docstring PEP, and it should also describe the how in more detail, so that I can understand what the function is doing without reading its code.

I also think that the function is too long.

Finally, `ENHYDRIS_TS_GRAPH_BIG_STEP_DENOMINATOR`, besides the fact it's a bad name, should be abolished as a setting. This thing shouldn't be configurable. If you want to be able to override it in the settings, you can make it a module or class variable.

enhydris/templates/enhydris/timeseries_detail/chartjs.html
```diff
@@ -6,96 +6,112 @@
 <script language="javascript" type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/flot/0.8.1/jquery.flot.selection.min.js"></script>
 <script language="javascript" type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/flot/0.8.1/jquery.flot.time.min.js"></script>
 <script type="text/javascript">
+// helper for returning the weekends in a period
```
aptiko, line 9: Is this comment in the correct position (or does it make any sense at all?)

Anyway, I'm not reviewing the rest of the js since my understanding is that it's only a few changes in the old code pending complete rewrite.

enhydris/api/tests/test_views/test_timeseries.py
```diff
@@ -520,3 +520,122 @@ def test_authorized_user_can_delete_timeseries(self):
...
+@patch("enhydris.models.Timeseries.get_data")
+class TimeseriesChartTestCase(APITestCase, TimeseriesDataMixin):
+    def create_timeseries(self):
+        # Create the timeseries so that we have 5 entries, one per year basically
+        super().create_timeseries()
+        self.htimeseries.data = pd.DataFrame(
+            index=[datetime(year, 1, 1) for year in range(2010, 2015)],
+            data={
+                "value": [year for year in range(2010, 2015)],
+                "flags": ["" for _ in range(5)],
```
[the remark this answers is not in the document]
AbdullahKady: The reasoning I had in mind for using comprehension was to make it obvious to the reader that it corresponds to the years. I had it as `["" for _ in range(2010, 2015)]` but for some reason I may have ended up changing it while working or maybe I just had that in mind and never actually wrote it?
Anyway, changed it to `[""] * 5`, listing them won't be suitable when it comes to `[""] * 10`

enhydris/api/tests/test_views/test_timeseries.py
```diff
@@ -520,3 +520,122 @@ def test_authorized_user_can_delete_timeseries(self):
...
+        self.assertEqual(response.status_code, 200)
+
+    def test_all_values_returned(self, mock):
+        mock.return_value = self.htimeseries
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        data = response.json()
+        self.assertEqual(len(data), 5)
+        expected_values = ["{}.00".format(y) for y in range(2010, 2015)]
+        self.assertEqual([x["value"] for x in data], expected_values)
```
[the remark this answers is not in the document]
AbdullahKady: Yes, I didn't use f-strings since I didn't encounter it in the codebase so I didn't want to introduce it.
Should I change it even in the `setUp` then (it won't be consistent with the rest of the test)

enhydris/api/tests/test_views/test_timeseries.py
```diff
@@ -520,3 +520,122 @@ def test_authorized_user_can_delete_timeseries(self):
...
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        data = response.json()
+        self.assertEqual(len(data), 4)
+        expected_values = ["{}.00".format(y) for y in range(2011, 2015)]
+        self.assertEqual([x["value"] for x in data], expected_values)
+
+    @override_settings(ENHYDRIS_TS_GRAPH_BIG_STEP_DENOMINATOR=3)
+    def test_data_sampled_by_equal_time_distance(self, mock):
+        self.htimeseries.data = pd.DataFrame(
```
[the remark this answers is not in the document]
AbdullahKady: I'm used to grouping Test Cases based on their functionality or their logical relation. Specially that this test case won't have -can't think of?- more than one test case, I didn't think it made sense to be separated.

enhydris/api/tests/test_views/test_timeseries.py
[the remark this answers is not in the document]
AbdullahKady: And I only opted out for testing the values rather than the objects/dicts of date-value since the dates won't be straight forward to test; seeing that we use timestamps and sample based on an equal distant, and 2012 being a leap year doesn't make it that simple.

enhydris/api/views.py
AbdullahKady: It's not actually wrong (of course `ie` is a mistake). I actually try to make the comments as short as possible (while obviously keeping the core words that explain it), and in "passive voice" whenever applicable.
So basically even after you pointed it out, I don't see the flaw with it to be honest. Anyway, I have changed them accordingly, just thought I would share my point of view :grin:

enhydris/api/views.py
AbdullahKady: Thank you, that was spot on!
I actually struggled a bit with the naming, but it was certainly because I was focusing on naming it based on the `how` rather than the `what`.
I have modified the comment(s), please check them and let me know since it seems that we have different views of "expressive comments", maybe due to our native tongues?

enhydris/templates/enhydris/timeseries_detail/chartjs.html
AbdullahKady: No it's not in the correct position, but it wasn't either before modifying the code actually :grin:, and I have no idea what it means in all honesty. I just removed it.

Also note that I added `// MODIFIED` comment to the lines that were actually changed, eventually this piece code should be completely thrown anyway so I thought it's okay

enhydris/api/tests/test_views/test_timeseries.py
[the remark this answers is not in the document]
aptiko: Then maybe you could introduce another method in that TestCase, maybe `_assertDictsAlmostEqual()`.

enhydris/api/tests/test_views/test_timeseries.py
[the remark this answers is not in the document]
aptiko: My opinion (but I'm not really an expert) is that Django gets unit testing wrong. I think that generally each test should have a single assertion. The way the Django testing framework is made I think it discourages this, for various reasons.

In many cases, rather than this:

    class SomeTestCase(TestCase):
        def test_something(self):
            do_something()
            self.assertEqual(a, b)  # (1)
            self.assertEqual(c, d)  # (2)

it may be better to write this:

    class SomeTestCase(TestCase):
        def setUp(self):
            do_something()

        def test_a_equals_b(self):
            self.assertEqual(a, b)

        def test_c_equals_d(self):
            self.assertEqual(c, d)

The reason for this is: how can you know that statement (2) works as intended? You can go in the main code and deactivate the functionality that statement (2) is intended to test, and then the test should fail. But if that functionality causes statements (1) and (2) to fail at the same time, then you'll never be able to tell whether (2) works, because it will never be triggered if it is to fail.

Now, in the second example above, `do_something()` is executed twice, once for each test. For tests that are so simple you don't want this, especially if `do_something()` takes a lot of time and the TestCase has many tests. What you can do is use `setUpClass()` instead. But in Django I think you can't do this, because the database preparation, truncation etc. happens just before and after `setUp()`. So you have to write tests as in the first example.

As I said, I'm not an expert, but apparently George Stocker, who has just started podcasting on these things, is. He told me that [in his second episode](https://www.buildbettersoftware.fm/2), released yesterday, he deals with the difference between TDD vs. just writing tests, and I think that this is very relevant to the above. (Unfortunately George Stocker is a .net guy.)

If you write tests like in the second example above, then your testing code becomes much larger in number of lines. I think that this is not a problem.

Back to our case: We have this case:

    class SomeTestCase(TestCase):
        def setUp(self):
            do_something()

        def test_1(self):
            test_something()

        def test_2(self):
            test_something_else()

        def test_3(self):
            ignore_setup_and_do_some_different_initialization()
            test_a_third_thing()

I also come across this problem sometimes, and sometimes I don't want to take out `test_3()` in its own class. What I do then is this:

    class SomeTestCase(TestCase):
        def _create_test_data(self, *args, **kwargs):
            do_something_with_args_and_kwargs()

        def test_1(self):
            self._create_test_data(...)
            test_something()

        def test_2(self):
            self._create_test_data(...)
            test_something_else()

        def test_3(self):
            self._create_test_data(...)
            test_a_third_thing()

Let me see if I can find an example where I did this... well, I can't find exactly this case, but a similar one is [`enhydris.tests.test_models.StationLastUpdateTestCase`](https://github.com/openmeteo/enhydris/blob/master/enhydris/tests/test_models.py#L295).

enhydris/api/views.py
aptiko: The flaws were
(1) "exceeds tolerance" What is the subject of "exceeds"?
(2) "a graceful null return": A graceful null return isn't "raised", so you can't really say "a KeyError is raised rather than a graceful null return", you need to say "... than a graceful null is returned" or something like that.

enhydris/api/tests/test_views/test_timeseries.py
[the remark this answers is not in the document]
gortok: @aptiko I spend most of my days on .NET; but still love python (I've produced commercial software using Python and Django, and I'd never say no to working with either again).

But, to hit on your point:

A test should have only one reason to fail.

If a Test can fail for more than one reason, then at some point, *it will*, and that'll cause confusion, and more insidiously, can cause your tests to become coupled to a particular implementation of a behavior, and that's not great.

What you're writing about testing one method fits into that.

This is incredibly hard to pull off when writing tests after you've written the code; and it's not easy even when you write the test first with TDD.  No one should feel bad for getting this wrong; I still do. Often.

enhydris/api/tests/test_views/test_timeseries.py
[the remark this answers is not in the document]
AbdullahKady: @aptiko added timestamp assertions, I had to create it as a mixin, however the inheritance is now causing the class definition to wrap, which to me -and I guess you'd agree- seems ugly.

I wanted to merge it into the `TimeseriesDataMixin` and call it `TimeseriesTestCaseMixin` or something along the lines, and the definition of an extra function that is not being used in all of the inheriting classes shouldn't be a problem I think, but I wanted to know your opinion.

enhydris/api/tests/test_views/test_timeseries.py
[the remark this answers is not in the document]
AbdullahKady: @aptiko and George -since I can't mention him-, thank you so much for the detailed fruitful discussion! :pray:

enhydris/api/tests/test_views/test_timeseries.py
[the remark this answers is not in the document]
aptiko: I do find it ugly, but I think there's no rule that says that classes shouldn't be inheriting from many classes. (Actually I think "multiple inheritance" is frowned upon, but this technique of calling one of the base classes "the base class" and naming the rest ones "mixins" works well I think.) Let me see what Robert Martin has to say about this... I don't think he says anything at all.

So I think it's fine to leave it like that.

enhydris/api/tests/test_views/test_timeseries.py
```diff
@@ -520,3 +520,155 @@ def test_authorized_user_can_delete_timeseries(self):
...
+        mock.return_value = self.htimeseries
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+
+    def test_all_values_returned(self, mock):
+        mock.return_value = self.htimeseries
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        data = response.json()
+        self.assertEqual(len(data), 5)
```
aptiko, line 626: These lines are repeated every time `_assertChartResponse()` is called. Could they be included in it?

```
self._assertChartResponse(response, expected)
```

enhydris/api/tests/test_views/test_timeseries.py
```diff
@@ -520,3 +520,155 @@ def test_authorized_user_can_delete_timeseries(self):
...
+            end_date=datetime(2017, 3, 1, 0, 0, tzinfo=self.tz.as_tzinfo),
+        )
+
+
+class TimeseriesChartTestMixin:
+    def _assertChartResponse(self, data, expected, tolerance_in_days=2):
+        """Assert chart response by allowing timestamp tolerance, but not values
+
+        The expected is a list of {value, date} rather than timestamp to make the
+        tests easier to write, they're translated into timestamps for comparision
```
aptiko, line 577: Add full-stops at the end of the heading and the text, fix the spelling of "comparison".

enhydris/api/tests/test_views/test_timeseries.py
AbdullahKady: Actually, I left them there intentionally. You can see that the `len` assertion is actually obsolete since we're testing the actual list anyway.
However I have had this habit from the main project that I started working on using Django for the first time, where the developers encouraged this to make the test sequence easily understandable for humans -aside from its name-, i.e. when I look at the test, I would go through the lines one-by-one to understand the scenario even if the assertion is not really useful, so we have 10 data points, and a max of 3 data points, even if we are asserting the actual 3 data points, then `len` tells me that beforehand, and an assertion about the status code tells me that the test is positive-assertion (non-error).

Not sure about that approach in all honesty, just thought I would share it anyway.

aptiko: OK, but I'm not certain this has much to do with what I meant. Why don't you change `_assertChartResponse()` to this:

    def _assertChartResponse(self, response, expected):
        self.assertEqual(response.status_code, 200)
        actual = response.json()
        self.assertEqual(len(actual), len(expected))
        # Plus whatever else _assertChartResponse() is already doing

This way you'll save three lines of code each time you call it.

AbdullahKady: @aptiko Yea I did that in my last commit: I moved the status code assertion there, but the asserting the length now makes no sense; since if I want it to be easily readable whenever the method is called, I need to pass it as an argument. If not, then I will need to that the `len(expected)` is equal to the length of `response.json()`, which doesn't really make sense since I'll assert them being equal anyway.

aptiko:
Merged in e8a60e8.

PR #323, opened 2020-07-19
Add time series groups (fixes #314)

### LINKED ISSUE #314: Add time series groups
For more information, see aptiko/openhi#77.

aptiko:


enhydris/models.py
```diff
@@ -377,6 +384,56 @@ class Timeseries(models.Model):
     )
     time_zone = models.ForeignKey(TimeZone, on_delete=models.CASCADE)
     remarks = models.TextField(blank=True)
+
+    def get_name(self):
```
AbdullahKady, line 388: suggestion: why not override the save/clean method -depending on the desired behavior- to populate the name with the variable if it's blank and omit this method.

enhydris/models.py
```diff
@@ -377,6 +384,56 @@ class Timeseries(models.Model):
...
+        if self.name:
+            return self.name
+        else:
+            return self.variable.descr
+
+    def __str__(self):
+        return self.get_name()
+
+    def get_default_timeseries(self):
```
AbdullahKady, line 397: I'm not sure how much this is used, so it might be over-optimization. This does -potentially- 3 queries, seeing that it's being used in the `start_date` and `end_date`, which I assume a view might display both very regularly, you can cache the result of `get_default_timeseries` or just use  `cached_property` as maybe `default_timeseries`

enhydris/models.py
```diff
@@ -377,6 +384,56 @@ class Timeseries(models.Model):
...
+    AGGREGATED = 400
+    TIMESERIES_TYPES = (
+        (RAW, _("Raw")),
+        (CHECKED, _("Checked")),
+        (REGULARIZED, _("Regularized")),
+        (AGGREGATED, _("Aggregated")),
+    )
+    last_modified = models.DateTimeField(default=now, null=True, editable=False)
+    timeseries_group = models.ForeignKey(TimeseriesGroup, on_delete=models.CASCADE)
+    type = models.PositiveSmallIntegerField(choices=TIMESERIES_TYPES)
```
AbdullahKady, line 436: Any reason why it's not a `CharField` with constants as `RAW = 'RAW'`? because this seems a bit confusing, and will be hard if I'm accessing the database directly outside from Django

enhydris/models.py
```diff
@@ -389,15 +446,33 @@ class Timeseries(models.Model):
     )
 
     class Meta:
-        verbose_name = "Time Series"
-        verbose_name_plural = "Time Series"
-        ordering = ("hidden",)
+        verbose_name = "Time series"
+        verbose_name_plural = "Time series"
+        ordering = ("type",)
```
AbdullahKady, line 451: I guess I understand the reasoning behind IntegerFields, but I don't actually see it used in the model (i.e. `self.timeseries_set.first()` for the `get_default_timeseries`)

enhydris/views.py
```diff
@@ -41,17 +42,49 @@ def get_redirect_url(self, *args, **kwargs):
...
+        timeseries_group_id = get_object_or_404(
+            models.Timeseries, id=timeseries_id
+        ).timeseries_group.id
+        station_id = get_object_or_404(
+            models.TimeseriesGroup, id=timeseries_group_id
+        ).gentity.id
         return reverse(
-            "timeseries_detail", kwargs={"station_id": station_id, "pk": timeseries_id}
+            "timeseries_group_detail",
+            kwargs={"station_id": station_id, "pk": timeseries_group_id},
```
AbdullahKady, line 70: why x2 `get_object_or_404` rather than 
```python
timeseries_group_id = get_object_or_404(
            models.Timeseries, id=timeseries_id
).timeseries_group
kwargs={"station_id": timeseries_group.gentity.id, "pk": timeseries_group.id},
```

enhydris/api/csv.py
```diff
@@ -79,8 +81,11 @@ def prepare_csv(queryset):
                 csvwriter = csv.writer(timeseries_csv)
                 csvwriter.writerow(_timeseries_list_csv_headers)
                 for station in queryset:
-                    for timeseries in station.timeseries.order_by("variable__id"):
-                        csvwriter.writerow(_timeseries_csv(timeseries))
+                    timeseries_groups = station.timeseriesgroup_set.order_by(
```
AbdullahKady, line 84: You can use `prefetch_related` with the custom ordering on the `queryset` to minimize the queries

enhydris/api/views.py
```diff
@@ -153,39 +155,47 @@ def create(self, request, *args, **kwargs):
...
         try:
-            gentity_id = int(serializer.get_initial()["gentity"])
+            timeseries_group_id = int(serializer.get_initial()["timeseries_group"])
         except ValueError:
             raise Http404
-        station = get_object_or_404(models.Station, id=gentity_id)
+        timeseries_group = get_object_or_404(
+            models.TimeseriesGroup, id=timeseries_group_id
+        )
+        station = get_object_or_404(models.Station, id=timeseries_group.gentity_id)
```
AbdullahKady, line 164: isn't the gentity a **required** `ForeignKey`? why not just access `timeseries_group.station` directly instead of 2 `get_object_or_404`?

enhydris/models.py
```diff
@@ -355,14 +355,21 @@ def path(self, name):
         return super().path(name)
 
 
-class Timeseries(models.Model):
+class TimeseriesGroup(models.Model):
     last_modified = models.DateTimeField(default=now, null=True, editable=False)
-    gentity = models.ForeignKey(
-        Gentity, related_name="timeseries", on_delete=models.CASCADE
-    )
+    gentity = models.ForeignKey(Gentity, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
aptiko: This line was in model `Timeseries`, and now it's in `TimeseriesGroup`. The related name would need to be changed to `timeseries_groups`. But I thought that there's no reason not to use the Django default, `timeseriesgroup_set`, so that the developer doesn't need to remember what the related name is in each case. I don't really remember why I had chosen `timeseries` instead of leaving the default `timeseries_set` before.

enhydris/models.py
aptiko: This looks complicated. E.g. what if the name of the variable changes? There are several edge cases and if we tried to be smart I think it would be confusing.

enhydris/models.py
aptiko: (See your other comment below.)

enhydris/models.py
aptiko: Yes, thanks for spotting that because I didn't think about it either. This affects the order with which the time series appear in the admin, so I'll leave it as it is for the time being.

enhydris/admin/station.py
```diff
@@ -120,6 +120,10 @@ def clean(self):
             self._check_submitted_data(self.cleaned_data["data"])
         return result
 
+    def clean_type(self):
+        return self.cleaned_data.get("type")
+        raise forms.ValidationError("You suck!")
```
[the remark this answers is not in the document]
aptiko: Fixed in 8d5f46f.

enhydris/models.py
aptiko: Fixed in a5a92a1.

enhydris/views.py
aptiko: Fixed in 72b8193.

enhydris/api/csv.py
aptiko: Fixed in dd9365c.

enhydris/api/views.py
aptiko: Fixed in 72b8193.

aptiko:
@AbdullahKady Thanks for the review. I fixed several things. While I agree that you don't need to look at the migration, I think it would be a good idea to look at the tests.

enhydris/models.py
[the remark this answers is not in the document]
AbdullahKady: Sounds good :+1:
I'm so used to defining a `related_name` whenever applicable, noticed that you usually omit them. Not really sure what the best practice is, will need to look it up.

enhydris/models.py
AbdullahKady: Not sure what the problem is with a name change? it will be re-saved as expected since it's not a blank value. Just a suggestion anyway :grin:

enhydris/api/views.py
```diff
@@ -153,39 +169,47 @@ def create(self, request, *args, **kwargs):
...
         if not request.user.has_perm("enhydris.change_station", station):
             return Response("Forbidden", status=status.HTTP_403_FORBIDDEN)
 
-        # Check the correctness of gentity_id
-        if str(gentity_id) != self.kwargs["station_id"]:
+        # Check whether the station and timeseries group ids in the url and in the
+        # posted data are consistent
+        if str(timeseries_group.gentity_id) != self.kwargs["station_id"]:
             return Response("Wrong gentity_id", status=status.HTTP_400_BAD_REQUEST)
+        if str(timeseries_group.id) != self.kwargs["timeseries_group_id"]:
```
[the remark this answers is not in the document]
aptiko: OK, I've re-made up my mind—I'm not writing the redirect view. First, loggertodb makes POST requests, and these can contain a large amount of data. I don't like making these requests twice. Second, writing that redirect view is some work, and I don't want to do it, as it would be short-lived. Third, no-one really uses the create view; in fact, no-one besides loggertodb is using that API. In the unlikely case that someone does make a create or other request and this results in an internal server error, it's no big deal.

Loggertodb is about to have similar changes and get a new version anyway, and after this is done and all users replace it, we'll remove the backwards compatible endpoint anyway.

enhydris/tests/test_models.py
```diff
@@ -376,44 +410,215 @@ def test_as_tzinfo(self):
...
+            type=type,
+            time_step="H",
+        )
+
+    def _test_str(self, type, result):
+        timeseries_group = mommy.make(models.TimeseriesGroup, name="Temperature")
+        timeseries = self._make_timeseries(timeseries_group, type)
+        self.assertEqual(str(timeseries), result)
+
+    def test_only_one_raw_per_group(self):
```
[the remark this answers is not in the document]
aptiko: Each of these tests tests a different `UniqueConstraint`, so I believe that each case should be a different test.

BTW, I took this opportunity to write https://wiki.antonischristofides.com/en/sops/production/development/writing-unit-tests.

PR #327, opened 2020-08-24
WIP: Use apex chart.

yantakus:


enhydris/templates/enhydris/timeseries_group_detail/chartjs.html
```diff
@@ -1,116 +1,233 @@
...
+      brush: {
+        target: "chart2",
+        enabled: true,
+      },
+      selection: {
+        enabled: true,
+      },
+      events: {
+        selection: function (chartContext, { xaxis }) {
+          debounceRefetch(xaxis);
```
AbdullahKady, line 166: I don't think it makes sense to use debounce here, I'd assume the library has a separate event for selection completed or something (i.e. when mouse is up), aside from performance, I don't think as UX it makes sense that the chart is actually changing when I haven't yet finished my selection range.

enhydris/templates/enhydris/timeseries_group_detail/chartjs.html
```diff
@@ -1,116 +1,233 @@
...
+      tickAmount: 2,
+    },
+  };
 
-  $('#overview').bind('plotselected', function (event, ranges) {
-    d = d_bak;
-    plot.setSelection(ranges);
-  });
-}
+  var chartLine = new ApexCharts(
```
AbdullahKady, line 198: Please rename them correctly, `chart` and `chartLine` is confusing, specially since `chartLine` is not actually a chart line, but rather an area one.
Maybe call it `rangeSelectionChart` or `miniRangeChart`.

enhydris/templates/enhydris/timeseries_group_detail/chartjs.html
```diff
@@ -1,116 +1,233 @@
...
-    // don't fire event on the overview to prevent eternal loop
-    overview.setSelection(ranges, true);
-  });
+  var optionsLine = {
+    series: [
+      {
+        data: [],
+      },
+    ],
+    chart: {
```
AbdullahKady, line 153: please add `noData` here as well, and actually if there are much shared options, please extract them in a shared variable and spread it for each one like so
```js
const sharedOptions = {
  noData: {text: "Loading..."}
};
const lineChartOptions = {...sharedOptions, newOption: "some-value"};
```

enhydris/templates/enhydris/timeseries_group_detail/chartjs.html
```diff
@@ -1,116 +1,233 @@
...
+        .substring(0, 16)}&end_date=${new Date(max * 1000)
+        .toISOString()
+        .substring(0, 16)}`
+    )
+      .then(function (response) {
+        return response.json();
+      })
+      .then(function (data) {
+        if (data && data[0]) {
+          const mappedData = data.map((i) => [
```
AbdullahKady, line 80: What's the max line length? I'd say follow python's configured one, which is 88, thus this shouldn't wrap.

AbdullahKady (changes requested):
- Panning doesn't actually work? so I think we need to remove its icon?
- The mini chart selected range isn't affected when the bigger chart limits are edited, it should be both ways (just like the old version).

yantakus:
@AbdullahKady what browsers are you supporting? I don't see any JS processing configured. So I just wanted to add fetch polyfill and it would support pretty old browsers, that's why I didn't use constants or arrow functions.

enhydris/templates/enhydris/timeseries_group_detail/chartjs.html
yantakus: No, it doesn't. The only option is to use debounce.

yantakus:
> * The mini chart selected range isn't affected when the bigger chart limits are edited, it should be both ways (just like the old version).

I've created an issue in apex charts repo, have a look: https://github.com/apexcharts/apexcharts.js/issues/1817

aptiko:
@yantakus @AbdullahKady We don't support old browsers. No IE.

Thanks for this work and the review.

@yantakus Sorry for being lazy and not wanting to go through the review discussion in detail. Where are we now? Have you addressed the review comments? (You don't need to do everything @AbdullahKady asked—the review is just so that an extra pair of eyes looks at this.) When it's ready, tell me so that I also take a quick look.

yantakus:
@aptiko Sorry for not responding for so long. Not it's clear regarding browser support requirements. Will be able to finish this PR on Monday.

leeoniya:
in case you guys want to consider a much smaller and faster alternative to Apex: https://github.com/leeoniya/uPlot#performance

yantakus:
@leeoniya Thanks, looks interesting. Is it possible to create such a chart with your library: https://openmeteo.org/stations/1468/timeseries/9630/?

leeoniya:
take a look at https://leeoniya.github.io/uPlot/demos/zoom-ranger.html

yantakus:
@aptiko it's very simple, but blazing fast. I'd try it out. Also, apexcharts seems to be dead, nobody cares about the issues posted in the repo. What do you think?

aptiko:
@yantakus I think I prefer keeping apexcharts. Here are the reasons:

1. Meanwhile I also used them in another project. What I found out was that although I needed something quite simple, I used a large number of features to make the chart show as the users needed it to in that particular use case. I feel safer using a mature library than a newer one that might not have the features I need.
2. Apexcharts doesn't seem to be dead. If you look at the list of commits there's considerable activity. They also have a support page where you can pay $30 or $50 and get support or have bugs fixed (btw if you need that you can use it and then tell me and I'll give you the amount).
3. We already made the decision to use apex charts. If we made this discussion two months ago, I'd say, by all means, yes, let's try uPlot. But now that we made the decision and we wrote a little code I think uPlot's smallness and fastness aren't reasons compelling enough to reconsider.

aptiko:
Superseded by #366.

PR #348, opened 2020-10-07
Fix access to empty timeseries group detail page (fix #341)

### LINKED ISSUE #341: Error on timeseries group detail if there's no time series
Create a time series group that has no time series; try to visit its detail page.

eyobofficial:
Closes #341


enhydris/templates/enhydris/timeseries_group_detail/main-default.html
```diff
@@ -20,7 +22,13 @@
       <div class="card-body">
         <div class="row">
           <div class="col-xs-12 col-sm-5">
-            {% include "enhydris/timeseries_group_detail/chart.html" %}
+            {% if object.timeseries_set.exits %}
+              {% include "enhydris/timeseries_group_detail/chart.html" %}
+            {% else %}
+              <div class="no-chart-label">
```
aptiko, line 28: I think it isn't clear whether this means "label for no chart", which is probably what you mean, or "no label for chart". Also I think that this isn't a label, it's a message. Maybe `message-no-data`?

enhydris/templates/enhydris/timeseries_group_detail/main-default.html
```diff
@@ -20,7 +22,13 @@
...
+                <h4 class="text-muted text-uppercase text-center">{% trans 'No Chart Data' %}</h4>
```
aptiko, line 29: I think this message is a bit misleading. It's not that there's no chart data, it's that there's no data. I'm thinking something more verbose, such as `This time series group has no data (yet)`. Since this is a longer message, it should probably be in normal size and without `text-uppercase`.

enhydris/templates/enhydris/timeseries_group_detail/download.html
```diff
@@ -4,60 +4,68 @@
...
-                  class="form-check-input"
-                  {% if forloop.first %}checked{% endif %}
-                >
-                <label
-                  class="form-check-label" 
-                  for="id_{{ download_data_form.timeseries_id.name }}_{{ forloop.counter0 }}"
-                >{{ choice }}
-                </label>
-              </div>
+    {% if object.timeseries_set.exists %}
```
aptiko, line 7: The contents of this `if` tag are too long and there's too many levels of indentation, so I think it may be hard to locate the corresponding `{% else %}` So maybe you could take the contents of the `{% if %}` and put them in another file that will be included. Read our [style guide on breaking large Django template files](https://wiki.antonischristofides.com/en/sops/production/development/code-style#breaking-large-files).

aptiko (changes requested):
Overall I like it. Except for my other comments, I think it would be good to unit test this. Not anything fancy; a single unit test with something like `self.assertContains(response, "whatever 'no data' message we have")` would probably suffice.

enhydris/templates/enhydris/timeseries_group_detail/main-default.html
eyobofficial: But that would mean we are using the term `time series group` for the end user. My understanding was that we should try to avoid using the term `time series group` from showing to the end user as much as possible.

aptiko: That's right. Yesterday I had a meeting with the hydrologists, and we discussed this again.

Technically the correct term is **variable**. For example, "wind speed" measured by a given sensor is a variable. The standard height for placing the wind sensor is 2m. If the sensor is placed 5m high, you can use a formula that can convert the 5-meter-high wind speed to the 2-meter-high wind speed. So if you process the raw wind speed **variable** you can create another **variable** (i.e. time series group), the "wind speed adjusted to 2m height". Or, another example, if you have two wind speed sensors, you have two variables (i.e. time series group), one is the wind speed measured by sensor X, the other is the wind speed measured by sensor Y.

Now, what do you call the concept of "wind speed" in general? Or "temperature"? Technically the correct term is **variable type**. So the "wind speed measured by sensor X" is a variable whose type is "wind speed".

Why didn't we call them thus then? The reason is that the users often use "variable" to refer to what we call "variable type". The problem is made worse because, although cases like the above examples exist (these are real examples from our database), in most cases each station has only one variable for each variable type, i.e. in most cases a station will have a single temperature time series group, a single humidity time series group, and so on. We concluded that the level of confusion would be unacceptable if we used "variable" for the time series group and "variable type" for what we have been calling a "variable" for at least the last 10 years.

So I asked them could we use "parameter" for the time series group? They clearly said no—it's inaccurate and it isn't going to make things better.

But they also said: "Why don't you like 'time series group'"? Apparently they like it and they aren't confused by it. They're less reluctant to show the term to the user than I am. So I thought that maybe it doesn't harm to start relaxing the rule you've mentioned.

In addition, very few people should actually see this message, because the time series group should start getting some data soon afterwards.

But if you think of a better message, that's fine.

eyobofficial: I think the reasoning makes sense and gives a more clearer picture. In that case, I think it's better to use the message you suggested as it's straigth-forward.

aptiko:
Thanks! Can I ask you for two more stupid things?

1. Could you remove trailing spaces from lines 33 and 52 of `button-active.html`?
2. I think it would be better to rename directory `download-button-partials` to `download_button` (i.e. two changes: the `partials` goes away and the hyphen becomes an underscore). Likewise, `button-active.html` and `button-disabled.html` should be with an underscore. (I have no opinion on whether an underscore is better than a hyphen here, but since so far we're apparently using underscores let's make it consistent.)

eyobofficial:
> Thanks! Can I ask you for two more stupid things?
> 
>     1. Could you remove trailing spaces from lines 33 and 52 of `button-active.html`?
> 
>     2. I think it would be better to rename directory `download-button-partials` to `download_button` (i.e. two changes: the `partials` goes away and the hyphen becomes an underscore). Likewise, `button-active.html` and `button-disabled.html` should be with an underscore. (I have no opinion on whether an underscore is better than a hyphen here, but since so far we're apparently using underscores let's make it consistent.)

That's not stupid at all. I will make the changes right away. In the past, I have used the [editorConfig](https://editorconfig.org/) library to create consistent formatting (tabs vs spaces, tab sizes, trailing lines, etc) across multiple IDEs. I find it to be helpful during collabration with other developers who use different types of IDEs. In the future, we could also consider that.

aptiko:
Thanks for telling me about editorConfig. For Python we have black and for JS we have eslint, so it would be useful only for Django templates. I'd prefer a tool that makes a more extensive check than just looking at the indentation and spaces, but I doubt it exists.

eyobofficial:
> Thanks for telling me about editorConfig. For Python we have black and for JS we have eslint, so it would be useful only for Django templates. I'd prefer a tool that makes a more extensive check than just looking at the indentation and spaces, but I doubt it exists.

editorConfig library is not used for linting purposes. It plays a different role. It's meant for creating a similar IDE enviroment for developers during development. For example, in the `.editorConfig` configuration file,  lets say we have stated the recommended tabsize for Python files to be 4 and for HTML, CSS, and JS files to be 2. Now, no matter which IDE the developer is using (VSCode, Sublime, Brackes, Atom, etc), when the developer uses the tab key, the IDE will automatically adjust the tabsize depending on the file type defined in the `.editorConfig` file. That way, it assist the developer to follow the same consistent style without worrying about their IDE.
But once the code is finished, it still has to be linted using the Black & elint before it can merged to the main branch of the repository. We don't have to use it if you think that's not that useful in this case but I feel like I have to clarify the difference.

Issue #351, opened 2020-10-13
Always shows the message "There is no data in this time series group".


aptiko:
Regardless whether there is or there isn't data, it always shows no chart and the "no data" message.

PR #356, opened 2020-10-14
Add unit test for the fix by issue #351 (fixes #355)

### LINKED ISSUE #355: Unit test the fix of #351
I merged the fix for #351 because I was in a hurry to install it on production, which was broken; however the fix should be unit-tested.

eyobofficial:
Closes #355

### DIFF
--- a/enhydris/tests/test_views.py
+++ b/enhydris/tests/test_views.py
@@ -411,6 +411,9 @@ class TimeseriesGroupDetailTestCase(TestCase, TimeseriesDataMixin):
         )
         self.assertContains(self.response, "This time series group has no data yet.")
 
+    def test_timeseries_group_with_timeseries(self):
+        self.assertNotContains(self.response, "This time series group has no data yet.")
+
     def test_title(self):
         self.assertContains(
             self.response, "<title>Beauty — Komboti — Enhydris</title>", html=True
--- a/enhydris/tests/test_views.py
+++ b/enhydris/tests/test_views.py
@@ -409,10 +409,18 @@ class TimeseriesGroupDetailTestCase(TestCase, TimeseriesDataMixin):
         self.response = self.client.get(
             f"/stations/{self.station.id}/timeseriesgroups/{self.timeseries_group.id}/"
         )
-        self.assertContains(self.response, "This time series group has no data yet.")
+        self.assertNotContains(self.response, "data_holder")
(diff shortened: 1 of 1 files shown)


aptiko (changes requested):
This unit test is very easy to break. If we change the message, it's almost certain we'll forget to change the unit test, and it will always be succeeding regardless whether the thing it's supposed to be testing works or not.

I think it would be better to change it with `assertContains(self.response, "Something that only exists when there's data")`.

eyobofficial:
@aptiko I am not exactly confident if my updated changes are also the right approach for the unit tests. So if you think this should also be changed (such as removing the HTML class & id names from the tests), let me know your comment on your feedback.

enhydris/tests/test_views.py
```diff
@@ -409,7 +409,18 @@ def test_timeseries_group_without_timeseries(self):
         self.response = self.client.get(
             f"/stations/{self.station.id}/timeseriesgroups/{self.timeseries_group.id}/"
         )
-        self.assertContains(self.response, "This time series group has no data yet.")
+        self.assertNotContains(self.response, "data_holder")
+        self.assertContains(self.response, "message-no-data")
+
+    def test_timeseries_group_with_timeseries(self):
+        self.assertContains(self.response, "data_holder")
+        self.assertNotContains(self.response, "message-no-data")
```
aptiko, line 417: I like this solution, and the symmetry between the tests.

enhydris/tests/test_views.py
```diff
@@ -409,7 +409,18 @@ def test_timeseries_group_without_timeseries(self):
...
+        self.assertContains(
+            self.response, f"{self.timeseries_group.start_date_naive:%Y/%m/%d %H:%M}"
+        )
+        self.assertContains(
+            self.response, f"{self.timeseries_group.end_date_naive:%Y/%m/%d %H:%M}"
+        )
```
aptiko, line 423: I think these additional asserts aren't necessary, and they violate the rule that each test should test only one thing. I'd just delete them.

BTW, now would be a good time to read "Writing unit tests" in the wiki.

aptiko (changes requested):

aptiko:
Merged in 97de984823b43cdaf3894ecc2e3f2fb9be2c1a38.

PR #366, opened 2020-11-02
Rewrite chart using Apex chart library

### LINKED ISSUE #316: Rewrite chart front-end code
The JS for the chart is in a mess and must be rewritten from scratch.

While we are currently using flot, we don't have a preference for it (neither are we against it). The person who decided to use flot and wrote the unreadable code isn't working with us any more. No-one in our team has any experience either with flot or with any other charting tool. Therefore, since the code is going to be rewritten anyway, we can use any charting tool. Proposals?

eyobofficial:
Closes #316


aptiko:
Thanks.

@eyobofficial Here's one problem it has. If you zoom into it, it's not granular enough—it seems to have one point every week or so. Each time you zoom in (or zoom out I guess) it should make a new request and bring in more granular data. AFAICS it doesn't make a new request on zoom in.

Now you'll notice that production has the same problem, although it *does* make a request when you zoom in. I don't know why this is. It used to work correctly in the past. I don't know when it broke, though I suspect e8a60e8.

While you take a look at this I'll add some infrastructure for linting and testing js.

aptiko:
Also the JS shouldn't be in the html, it should go in `enhydris.js`.

eyobofficial:
> Also the JS shouldn't be in the html, it should go in `enhydris.js`.

The problem with moving the code to a dedicated JS file is it would be impossible (as far as I know) to dynamically build the API URL for the current page. Is there any technic I am missing that can be used to solve that in a JS file?

```
  const config = {
    blue: "#008FFB",
    gray: "#546E7A",
    api: "{% url 'timeseries-chart' station_id=object.gentity.id timeseries_group_id=object.id pk=object.default_timeseries.id %}"
  };
```

aptiko:
Yes, we have this in `templates/enhydris/base/default.html` (which is the main template):
```
      <script>
        var enhydris = {};
        enhydris.staticUrl = '{% static "" %}';
        enhydris.rootUrl = '{% url "station_list" %}';
      </script>
      <script src="{% static 'js/enhydris.js' %}"></script>
```
So you can add more variables there.

eyobofficial:
> Thanks.
> 
> @eyobofficial Here's one problem it has. If you zoom into it, it's not granular enough—it seems to have one point every week or so. Each time you zoom in (or zoom out I guess) it should make a new request and bring in more granular data. AFAICS it doesn't make a new request on zoom in.
> 
> Now you'll notice that production has the same problem, although it _does_ make a request when you zoom in. I don't know why this is. It used to work correctly in the past. I don't know when it broke, though I suspect [e8a60e8](https://github.com/openmeteo/enhydris/commit/e8a60e8b0634cbc0c52aed27f8e5a5756dfc5097).
> 
> While you take a look at this I'll add some infrastructure for linting and testing js.

It seems to me all the points are getting plotted to the chart during the initial load. The poins are retrieved using the [charts endpoint](https://openmeteo.org/api/stations/1403/timeseriesgroups/476/timeseries/9505/chart/) that looks like it's returning all the associated timeseries values. Thus, when zooming in, there are no more granular data to retrieve that have not been already plotted. So my question is, is there other 'hidden' in-between values that are not getting returned by above endpoint? If so, is there a separate endpoint or some kind of optional query parameters to retrieve those in-between values.

Also, I couldn't find the timeseries values in the admin page to easily change and debug them. I am currently using the django shell to work with those data and that's a bit of a slow process. Could you provide me a link to the responsible admin page? Thanks!

aptiko:
You can't download time series data from the admin page, you need to go to the time series detail page (choose CSV format). You can only upload time series data from the admin page, by supplying a file in similar format to the one you donwloaded.

For the API end point, see also https://enhydris.readthedocs.io/en/latest/dev/webservice-api.html#time-series-chart-data.

aptiko (changes requested):

enhydris/static/js/chart.js
```diff
@@ -0,0 +1,220 @@
+$(function() {
+	// Global configuration values
+	const config = {
+		blue: "#008FFB",
+		gray: "#546E7A",
+	};
+
+	// Debounce function
+	const debounceRefetch = debounce(refetch, 200);
```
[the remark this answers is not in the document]
eyobofficial: 1. When the user selects two data points in order to zoom the chart, the built-in `fetch` function is called between these new data points (i.e. start datetime and end datetime) to retrieve more granular timeseries data and populate them on the chart. The `refetch` function encapsulates this functionality. To make this more clear, I have now changed the function name to `refetchChart`.

2. The way the `selection` event of the apex chart works is -- as soon as the user starts the selection process, the event is executed multiple times for each mouse drag. Since the `refetchChart` method is called by this event (and since the `refetchChart` method makes API calls to retrieve granular chart data), multiple API calls are made before the user is done selecting. To mitigate this, the `debounce` function takes a callable and returns another modified callable that is only executed after a certain amount of time. This way the generated callable from the `debounce` function gives the user a certain amount time (200 milliseconds in this case) before firing the `selection` event. You can checkout [this](https://www.freecodecamp.org/news/debounce-explained-how-to-make-your-javascript-wait-for-your-user-to-finish-typing-2/) and [this](https://gist.github.com/nmsdvid/8807205) links to read more about this pattern.

While the `debounce` function name is confusing, I avoided using another name as it is a commonly used name for this pattern in the JavaScript community and this might come handy in the future when looking to debug it. However, if you still think it's still better to rename it to something more fitting to our case, please let me know and I will quickly change it. 

**Note:** I have made some little modifications to the `debounce` function to make it ES6 compliant.

enhydris/static/js/chart.js
```diff
@@ -0,0 +1,220 @@
...
+
+	// Line chart
+	const optionsLine = {
```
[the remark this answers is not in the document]
eyobofficial: Instead of `largeChart` and `smallChart`, I now used  `mainChart` and `miniChart`. I felt that is more suitable for the case. But if you still think that the 'small' and 'large' combination is better, please let me know and I will quickly change it.

enhydris/static/js/chart.js
```diff
@@ -0,0 +1,236 @@
+$(function() {
+	// Global configuration values
```
aptiko, line 2: I think this comment isn't needed, things are clear enough.

enhydris/static/js/chart.js
```diff
@@ -0,0 +1,236 @@
...
+	mouse drags. Since the `refetchChart` function is part of the `selection`
+	event, this causes multiple API calls to the chart endpoint before the user
+	has completed the selection. This is not efficient. Thus, by using the
+	`debounce` function, we can create another function that delay the execution
+	of the `refetchChart` function by a certain amount of milliseconds, in order
+	to give the user enough time to complete the selection. 
+	*/
+	const debounceRefetchChart = debounce(refetchChart, 200);
+
+	// Main ("big") chart
```
aptiko, line 19: I think that "main chart" and "mini chart" are clear names, so this comment isn't needed.

enhydris/static/js/chart.js
```diff
@@ -0,0 +1,236 @@
...
+		}
+	};
+
+	const mainChart = new ApexCharts(
+		document.querySelector("#chart-main"),
+		mainChartOption
+	);
+	mainChart.render();
+
+	// Mini ("small") chart
```
aptiko, line 94: Likewise for this comment.

enhydris/static/js/chart.js
```diff
@@ -0,0 +1,236 @@
...
+		},
+	};
+
+	const miniChart = new ApexCharts(
+		document.querySelector("#chart-mini"),
+		miniChartOptions
+	);
+	miniChart.render();
+
+	// Make API call
```
aptiko, line 150: I think this comment isn't useful, it's pretty much clear that `fetch(enhydris.chartApiUrl)` is making an API call.

enhydris/templates/enhydris/timeseries_group_detail/main-default.html
```diff
@@ -53,3 +47,21 @@
...
+  <script>
+    enhydris.strLoading = "{% trans 'Loading...'  %}";
+    enhydris.strNoData = "{% trans 'No data' %}";
+
+    {% if object.default_timeseries %}
+      enhydris.chartApiUrl = "{% url 'timeseries-chart' station_id=object.gentity.id timeseries_group_id=object.id pk=object.default_timeseries.id %}";
+    {% endif %}
+  </script>
+
+  {# Only load Apex library & JavaScript code if chart data exists #}
```
aptiko, line 62: I think that this comment and the `{% if %}` that follows it is too much trouble for what they achieve. The JS is likely to be cached anyway, and if it isn't, it will be cached and used on a subsequent visit to another time series group, and if the user just happened to pass buy, this is his first such visit, and after finding out this time series group has no data he leaves and never comes back, well, he will have wasted a little bit of bandwidth (and no time, since this will probably not delay rendering at all). But if you have a different opinion it's ok to leave it as it is.

aptiko (changes requested):
Thanks! I think this is fine. I did make a few minor comments but this is because I'm very picky, it would be fine as is.

enhydris/templates/enhydris/timeseries_group_detail/main-default.html
eyobofficial: > I think that this comment and the `{% if %}` that follows it is too much trouble for what they achieve. The JS is likely to be cached anyway, and if it isn't, it will be cached and used on a subsequent visit to another time series group, and if the user just happened to pass buy, this is his first such visit, and after finding out this time series group has no data he leaves and never comes back, well, he will have wasted a little bit of bandwidth (and no time, since this will probably not delay rendering at all). But if you have a different opinion it's ok to leave it as it is.

I agree with the explanation. So I will make the changes accordingly.

enhydris/templates/enhydris/timeseries_group_detail/main-default.html
```diff
@@ -58,10 +58,8 @@
       enhydris.chartApiUrl = "{% url 'timeseries-chart' station_id=object.gentity.id timeseries_group_id=object.id pk=object.default_timeseries.id %}";
     {% endif %}
   </script>
-
-  {# Only load Apex library & JavaScript code if chart data exists #}
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/apexcharts/3.19.3/apexcharts.min.js"></script>
   {% if object.default_timeseries %}
```
aptiko, line 62: Shouldn't the `chart.js` also be loaded unconditionally?

aptiko (changes requested):

enhydris/templates/enhydris/timeseries_group_detail/main-default.html
eyobofficial: The `chart.js` makes an API call to the `chartApiUrl` endpoint. However, the `chartApiUrl` variable is constructed conditionally on Line 58 (otherwise constructing the URL would fail since `object.default_timeseries` is empty). Thus, if the `chart.js` is loaded without the `chartApiUrl`, the JS `fetch` function would break. This is also causing the unit tests to fail.

aptiko: OK, normally the template shouldn't have this kind of logic, but we'll keep it like this and fix it later. Our JS generally needs a little overhaul, so we'll proceed with setting up linting and unit testing and we'll fix it right afterwards.

aptiko:
Merged in 095f140553c5650cb8be885fbbb98c3e904232cb and ee69d07f848b855305635f90dad2f1a557997ba4.

PR #380, opened 2020-11-18
Add automated JS linting & testing infrastructure (fixes #367)

### LINKED ISSUE #367: Lint and test the JS
Currently our JS is unlinted, and we only have a few Selenium tests. Since Selenium is a pain, we want to avoid it to the extent possible so we'll start testing our JS with jest and mocking. We'll do what we also do in `openmeteo/aira`; so from there you need to copy `.eslintrc`, parts of `.travis.yml`, and most of `package.json`. If `openhi-docker-dev` needs changes (which it definitely does), tell me about them so that I fix it.

This issue is not about writing unit tests, but only for creating the infrastructure to write unit tests. So just write a trivial/dummy test (otherwise jest will complain that there's an error). You can get ideas from `openmeteo/aira`, from the files that end in `.test.js`.

eyobofficial:
Closes #367

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,4 @@
+.gitignore
+enhydris/static/js/vendor
+phantomjs*
+enhydris/static/js/arg.js.v1.1.min.js
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,16 @@
+{
+    "extends": "airbnb-base",
+    "env": {
+        "browser": true,
+        "jquery": true,
+        "jest": true
+    },
+    "globals": {
+        "enhydris": "writable",
+        "OpenLayers": "readonly",
+        "google": "readonly",
+        "moment": "readonly",
+        "L": "readonly",
+        "ApexCharts": "readonly"
+    }
+}
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,5 +12,9 @@
         "moment": "readonly",
         "L": "readonly",
         "ApexCharts": "readonly"
+    },
+    "rules": {
+        "prefer-arrow-callback": "off",
+        "func-names": "off"
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.pyc
 *.egg
 *.egg-info
+node_modules
 doc/_build
 dist
 timeseries_data
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ matrix:
     - name: "Python 3.7 unit tests"
       language: python
       python: 3.7
+      node_js: "10"
       cache: pip
       addons:
         postgresql: "10"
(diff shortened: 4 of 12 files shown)


aptiko (changes requested):
Thanks! Please fix the js files according to what the linter found. (It will fix many of these problems itself with `npm run lint --fix` I think.)

eyobofficial:
> Thanks! Please fix the js files according to what the linter found. (It will fix many of these problems itself with `npm run lint --fix` I think.)

@aptiko One of the errors flag by the linter is `no-underscore-dangle` in those variable names which are prefixed by an underscore. My understanding is that you have intentionally named some of the variables this way to communicate their private scope. Since this is done this way by design, I was thinking of updating the rules in the `.eslintrc` configuration file to ignore the `no-underscore-dangle` error when linting. Do you think this is a good idea? Or should I just go on and fix it as the default linter configuration suggested?

aptiko:
@eyobofficial Good question. My opinion is to fix it as the linter suggests.

Remember I'm not a JS expert. Naturally when I program in JS I'm bringing Python conventions in, and I thought that using a leading underscore to indicate a kind of "private" (or protected) variable is a good idea. However, if the JS community (at least the part of it that uses the airbnb JS convention) does not use this practice, then we shouldn't be using it—our JS code should appeal to JS programmers, not to Python programmers. Assuming that the linter configuration accurately reflects the community's practices, we should do what it says.

But it could be that you are more experienced in JS than I am, so if you think I've understood something wrong, I'm all ears.

eyobofficial:
> @eyobofficial Good question. My opinion is to fix it as the linter suggests.
> 
> Remember I'm not a JS expert. Naturally when I program in JS I'm bringing Python conventions in, and I thought that using a leading underscore to indicate a kind of "private" (or protected) variable is a good idea. However, if the JS community (at least the part of it that uses the airbnb JS convention) does not use this practice, then we shouldn't be using it—our JS code should appeal to JS programmers, not to Python programmers. Assuming that the linter configuration accurately reflects the community's practices, we should do what it says.
> 
> But it could be that you are more experienced in JS than I am, so if you think I've understood something wrong, I'm all ears.

It's true JS normally doesn't use leading underscores in its variables like Python does. But I wanted to make sure with you in case you have specific reasons. So I think I will go with the default suggestions of the linter in this case as per your advice. Thanks!

eyobofficial:
On the `enhydris.js` file (Line 43):
```
        var queryParams = Arg.all();
```
@aptiko ESLint is complaining that this Arg variable is not defined. And I couldn't trace where it came from or what it is doing. Could you explain to me what I seem to be missing so I could initialize or declare it? Thanks!

aptiko:
@eyobofficial This appears to be from https://github.com/stretchr/arg.js/. `enhydris/templates/enhydris/base/main_with_map-default.html` is the file that includes `args.js`.

aptiko:
Merged in f68cb1189e9dfb910e35923ec980c9ce4b0d779d.

PR #388, opened 2020-12-02
Cache date property values (fixes #384)

### LINKED ISSUE #384: Cache start_date, end_date and last_update (and the _naive counterparts)
`Station` has `last_update`, and `TimeseriesGroup` and `Timeseries` have `start_date` and `end_date`, and all these have a `_naive` version (total 10 properties). The front page, which fetches the `last_update` for many stations, could be quite expensive because it has to query a large number of time series. We should cache these in the Django cache.

All these caches should be invalidated in `Timeseries.save()`.

eyobofficial:
Closes #384

### DIFF
--- a/enhydris/models.py
+++ b/enhydris/models.py
@@ -314,14 +314,14 @@ class Station(Gpoint):
         verbose_name = _("Station")
         verbose_name_plural = _("Stations")
 
-    @property
+    @cached_property
     def last_update_naive(self):
         result = self.last_update
         if result is not None:
             result = result.replace(tzinfo=None)
         return result
 
-    @property
+    @cached_property
     def last_update(self):
         timeseries = Timeseries.objects.filter(timeseries_group__gentity_id=self.id)
         result = None
@@ -520,23 +520,23 @@ class TimeseriesGroup(models.Model):
             if timeseries.type == type:
                 return timeseries
 
-    @property
+    @cached_property
     def start_date(self):
         if self.default_timeseries:
             return self.default_timeseries.start_date
 
-    @property
+    @cached_property
     def end_date(self):
         if self.default_timeseries:
             return self.default_timeseries.end_date
 
-    @property
+    @cached_property
     def start_date_naive(self):
         result = self.start_date
         if result is not None:
             return result.replace(tzinfo=None)
 
-    @property
+    @cached_property
     def end_date_naive(self):
         result = self.end_date
         if result is not None:
@@ -596,7 +596,7 @@ class Timeseries(models.Model):
             ),
         ]
 
-    @property
+    @cached_property
     def start_date(self):
         try:
             return self.timeseriesrecord_set.earliest().timestamp.astimezone(
@@ -605,7 +605,7 @@ class Timeseries(models.Model):
         except TimeseriesRecord.DoesNotExist:
             return None
 
-    @property
+    @cached_property
     def end_date(self):
         try:
             return self.timeseriesrecord_set.latest().timestamp.astimezone(
@@ -614,7 +614,7 @@ class Timeseries(models.Model):
         except TimeseriesRecord.DoesNotExist:
             return None
 
-    @property
+    @cached_property
     def start_date_naive(self):
         try:
             return (
@@ -625,7 +625,7 @@ class Timeseries(models.Model):
         except TimeseriesRecord.DoesNotExist:
             return None
 
-    @property
+    @cached_property
     def end_date_naive(self):
         try:
             return (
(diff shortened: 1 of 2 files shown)


aptiko (changes requested):
I think that `@cached_property` is not the right way to do it. This cache will persist only as far as that instance exists. After a request completes, the object no longer exists, and neither does its cached property. In the next request it will hit the database again.

So, for example, instead of this that you have in a test:
```
with self.assertNumQueries(0):
    self.timeseries_group.start_date
```
try this:
```
with self.assertNumQueries(0):
    models.TimeseriesGroup.get(id=self.timeseries_group.id).start_date
```
and it will probably fail, because you are creating a new instance.

I believe you need to [use the Django cache explicitly](https://docs.djangoproject.com/en/3.1/topics/cache/#basic-usage), as we do with `Timeseries.get_data()`.

eyobofficial:
> I think that `@cached_property` is not the right way to do it. This cache will persist only as far as that instance exists. After a request completes, the object no longer exists, and neither does its cached property. In the next request it will hit the database again.
> 
> So, for example, instead of this that you have in a test:
> 
> ```
> with self.assertNumQueries(0):
>     self.timeseries_group.start_date
> ```
> 
> try this:
> 
> ```
> with self.assertNumQueries(0):
>     models.TimeseriesGroup.get(id=self.timeseries_group.id).start_date
> ```
> 
> and it will probably fail, because you are creating a new instance.
> 
> I believe you need to [use the Django cache explicitly](https://docs.djangoproject.com/en/3.1/topics/cache/#basic-usage), as we do with `Timeseries.get_data()`.

Ok, that's will make the changes to reflect that.

aptiko (commented):
Since this is great, we can waste some time and discuss about the very minor things :-)

You've used some functions in functions, like this:
```
@property
def last_update(self):
    def get_last_update():
        # ...

    cache.get_or_set("cache_key", get_last_update)
```
Here is how I'd write this:
```
@property
def last_update(self):
    cache.get_or_set("cache_key", self._get_last_update)

def _get_last_update(...):
    # ...
```
(Not certain whether it works, I haven't tested it.)

Although I avoid functions in functions, I understand how in this particular case they present the whole logic nicely. I'm thinking out loud here: According to Robert C. Martin, functions should be up to five lines large, which I try to do but I'm not very fanatic about it. So using a nested function makes the outside function already exceed that. If the inside function has the tendency to grow a bit, and you want to break it in more functions, you'll have to un-nest it, otherwise it will become too hard to read. So I think that my version is more general. I also think I prefer it because of the less indentation.

But you don't need to change it, I'm just exchanging views on clean code. I'm interested in your opinion.

aptiko:
Merged in 70a82c88e11e1649e2a8282fd09bb3ed6a4033a2.

BTW, your name in this commit appears as "eyob tariku". Please capitalize it properly.

eyobofficial:
> Since this is great, we can waste some time and discuss about the very minor things :-)
> 
> You've used some functions in functions, like this:
> 
> ```
> @property
> def last_update(self):
>     def get_last_update():
>         # ...
> 
>     cache.get_or_set("cache_key", get_last_update)
> ```
> 
> Here is how I'd write this:
> 
> ```
> @property
> def last_update(self):
>     cache.get_or_set("cache_key", self._get_last_update)
> 
> def _get_last_update(...):
>     # ...
> ```
> 
> (Not certain whether it works, I haven't tested it.)
> 
> Although I avoid functions in functions, I understand how in this particular case they present the whole logic nicely. I'm thinking out loud here: According to Robert C. Martin, functions should be up to five lines large, which I try to do but I'm not very fanatic about it. So using a nested function makes the outside function already exceed that. If the inside function has the tendency to grow a bit, and you want to break it in more functions, you'll have to un-nest it, otherwise it will become too hard to read. So I think that my version is more general. I also think I prefer it because of the less indentation.
> 
> But you don't need to change it, I'm just exchanging views on clean code. I'm interested in your opinion.

@aptiko Even though I am still going through Robert C. Martin's book, I understand your point. I am also usually reluctant to use nested functions and my first instinct was to use separate private methods as you suggested. However, the 'inner' functions are only used by the 'parent' functions. Thus, I first tried to use the `lambda` expression as the callable argument of the `get_or_set` method, but that obviously didn't work since Python's `lambda` expression can only take one line of expressions. So, I finally decided to use the nested functions so as not to 'pollute' the class with methods that are only relevant for specific methods. However, this is not a pattern I expect to use more often.

eyobofficial:
> Since this is great, we can waste some time and discuss about the very minor things :-)
> 
> You've used some functions in functions, like this:
> 
> ```
> @property
> def last_update(self):
>     def get_last_update():
>         # ...
> 
>     cache.get_or_set("cache_key", get_last_update)
> ```
> 
> Here is how I'd write this:
> 
> ```
> @property
> def last_update(self):
>     cache.get_or_set("cache_key", self._get_last_update)
> 
> def _get_last_update(...):
>     # ...
> ```
> 
> (Not certain whether it works, I haven't tested it.)
> 
> Although I avoid functions in functions, I understand how in this particular case they present the whole logic nicely. I'm thinking out loud here: According to Robert C. Martin, functions should be up to five lines large, which I try to do but I'm not very fanatic about it. So using a nested function makes the outside function already exceed that. If the inside function has the tendency to grow a bit, and you want to break it in more functions, you'll have to un-nest it, otherwise it will become too hard to read. So I think that my version is more general. I also think I prefer it because of the less indentation.
> 
> But you don't need to change it, I'm just exchanging views on clean code. I'm interested in your opinion.

@aptiko Even though I am still going through Robert C. Martin's book, I understand your point. I am also usually reluctant to use nested functions and my first instinct was to use separate private methods as you suggested. However, the 'inner' functions are only used by the 'parent' functions. Thus, I first tried to use the `lambda` expression as the callable argument of the `get_or_set` method, but that obviously didn't work since Python's `lambda` expression can only take one line of expressions. So, I finally decided to use the nested functions so as not to 'pollute' the class with methods that are only relevant for specific methods. However, this is not a pattern I expect to use more often.

PR #394, opened 2020-12-11
Add `Arg` variable to `globals` in .eslintrc file

eyobofficial:
Currently, a magic comment is used in the  enhydris.js file to instruct ESLint to ignore the `Arg` global variable. Instead, add the `Arg` variable to `globals` section in the .eslintrc file.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,7 +11,8 @@
         "google": "readonly",
         "moment": "readonly",
         "L": "readonly",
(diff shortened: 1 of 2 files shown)


aptiko:
@eyobofficial Thanks! I deleted the long description of the commit and left only its summary. I think this fix was trivial and didn't need explaining. We should generally avoid explaining the obvious.

PR #395, opened 2020-12-12
Fix excess decimal digits in the y-axis of the chart (fixes #390)

### LINKED ISSUE #390: Sometimes the chart shows too many decimal digits
https://openmeteo.org/stations/1458/timeseriesgroups/6/

eyobofficial:
Closes #390

### DIFF
--- a/enhydris/static/js/chart.js
+++ b/enhydris/static/js/chart.js
@@ -122,6 +122,11 @@ $(function () {
     },
     yaxis: {
       tickAmount: 2,
+      labels: {
+        formatter(val) {
+          return val.toFixed(1);
+        },
+      },
     },
   };
 
@@ -190,6 +195,13 @@ $(function () {
         enabled: false,
       },
     },
+    yaxis: {
+      labels: {
+        formatter(val) {
+          return val.toFixed(1);
+        },
+      },
+    },
     tooltip: {
       enabled: true,
       x: {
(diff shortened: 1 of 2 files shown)


enhydris/static/js/chart.js
```diff
@@ -122,6 +122,11 @@ $(function () {
     },
     yaxis: {
       tickAmount: 2,
+      labels: {
+        formatter(val) {
+          return val.toFixed(1);
```
aptiko, line 127: Using one decimal digit here is suboptimal. You'd better use the [precision attribute](https://enhydris.readthedocs.io/en/latest/dev/database.html#enhydris.models.TimeseriesGroup.precision); something like
```
val.toFixed(precision >= 0 ? precision : 0)
```
This would still be suboptimal, but it will do for now. (A better solution would be to make sure that the ticks appear in nicer positions, e.g. for temperature they should be at 0,5, 10 degrees etc. and not in whatever random steps the chart decides—but let's leave this for the future.)

enhydris/static/js/chart.js
```diff
@@ -190,6 +195,13 @@ $(function () {
         enabled: false,
       },
     },
+    yaxis: {
+      labels: {
+        formatter(val) {
+          return val.toFixed(1);
```
aptiko, line 201: Likewise.

enhydris/static/js/chart.js
```diff
@@ -209,6 +221,7 @@ $(function () {
     .then((response) => response.json())
     .then((data) => {
       if (data && data[0]) {
+        console.log('Returned data: ', data);
```
aptiko, line 224: Is this a forgotten debugging statement?

PR #400, opened 2020-12-20
Refactor the chart.js code

### LINKED ISSUE #389: Refactor chart JS code
Refactor the JS code so that it is as the wiki says in "Code style", "vanilla Javascript". It's going to be different than the rest of the code, which is based on another pattern, but this shouldn't be a problem.

I don't think you'll need any kind of inheritance, just move the code to `enhydris.js` and group the functionality in an object instead of in a function.

While you are at it, could you write a few unit tests for it? Something very simple, like the first few tests of `kc-chart.test.js` in aira. You'll need to mock `ApexCharts` (see the top few lines of `kc-chart.test.js`).

You may also need to mock `fetch()`. I'm not sure how this is done. Do it in the simplest possible way. If a simple `fetch = jest.fn()` does it, do it; if this doesn't work, search the web, there's lots of information about that. Maybe use `fetch-mock` or `fetch-mock-jest` or `jest-fetch-mock`. Don't mock `fetch` in a complicated way so that it returns stuff that resembles real API responses. Just mock it so that the functions that call it can return without error.

We only want to check some really rudimentary stuff, like that functions have been called (`expect(...).toHaveBeenCalled()`). At this stage we don't really want to do exhaustive testing; instead, we want to slowly get in the habit of test our JS.

And, BTW, don't unit-test `debounce()` at all. This function should be in the JS "standard library" or whatever it's called, and since it isn't, we should be using a third-party library that has it, and if a good one doesn't exist, we should create one ourselves. In any case, leave it as it is for now.

eyobofficial:
Closes #389


aptiko:
@eyobofficial I have one remark on your code, which I thought was too minor so I merged it anyway. At some point you have `const self = this`. Now, as you know, I'm not really an expert in JS, but I think that the custom is to use `const that = this`, that is, use the name `that`, not `self`. I'm not really certain. I've definitely seen a book use `that` (I don't remember which), but I've never seen `self`.

aptiko (commented):
a

PR #427, opened 2021-02-18
Adjust search block due to feedback (fixes #424)

### LINKED ISSUE #424: Use less space for "search stations" and "list stations visible on map"
Users complain that the search field and the "list stations visible on map" take too much space and they prevent them from viewing the stations. As a first step, I propose:

1) to remove the text from the "list stations visible on map" button (it should appear on hover though) and leave only a symbol, and put it on the right of the search help button (which appears with a question mark).
2) to remove the word "search" and instead use nothing? a right arrow? a magnifying glass? (in the latter case I guess we'll need to remove the magnifying glass on the left.)
3) to remove the search field width slightly.

If you think these changes are a good idea, please go ahead and make them. Otherwise, let's discuss about them.

FYI, in https://system.openhi.net/, I've added another search field on the bottom right (this is specific to the OpenHI project so I've put it in the customized theme for OpenHI; it's not in Enhydris).

ad0v0:
**Checklist**:

* [x] All tests are passing
* [x] I have added any necessary documentation and/or release notes
* [x] I have followed the [commit message guidelines](https://wiki.antonischristofides.com/en/sops/production/development/git-and-github#committing-and-commit-messages) and the ["before submitting a pull request" guidelines](https://wiki.antonischristofides.com/en/sops/production/development/git-and-github#before-submitting-a-pull-request)
* [x] If I introduced (or abolished) any third-party library, I will write a short paragraph in this PR [according to the guidelines](https://wiki.antonischristofides.com/sops/production/development/third-party-libraries)
* [x] I have removed all obsoleted code

### DIFF
--- a/enhydris/static/css/enhydris.css
+++ b/enhydris/static/css/enhydris.css
@@ -341,13 +341,14 @@ h4.tab-no-data {
   position: relative;
   z-index: 990;
 
+  display: flex;
+  align-items: center;
   margin-top: 30px;
   margin-right: auto;
   margin-left: auto;
 }
 
 .page-map-full .form-search {
-  width: 100%;
   margin: 0 auto;
 }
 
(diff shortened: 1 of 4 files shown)


aptiko:
Merged in bc711734b6

PR #447, opened 2021-05-12
Implement redesign (fixes #433)

### LINKED ISSUE #433: Implement new design for time series group detail

ad0v0:
**Checklist**:

* [x] All tests are passing
* [x] I have added any necessary documentation and/or release notes
* [x] I have followed the [commit message guidelines](https://wiki.antonischristofides.com/en/sops/production/development/git-and-github#committing-and-commit-messages) and the ["before submitting a pull request" guidelines](https://wiki.antonischristofides.com/en/sops/production/development/git-and-github#before-submitting-a-pull-request)
* [x] If I introduced (or abolished) any third-party library, I will write a short paragraph in this PR [according to the guidelines](https://wiki.antonischristofides.com/sops/production/development/third-party-libraries)
* [x] I have removed all obsoleted code

### DIFF
--- a/enhydris/static/css/enhydris.css
+++ b/enhydris/static/css/enhydris.css
@@ -229,24 +229,7 @@ h4.tab-no-data {
 }
 
 .breadcrumbs li:last-child a {
-  position: relative;
-
   color: #333333;
-}
-
-.breadcrumbs li:last-child a::after {
-  position: absolute;
-  top: 50%;
-  right: -15px;
-  content: "";
-
-  width: 10px;
-  height: 5px;
-  margin-top: -2.5px;
-
-  background-image: url("../images/icon-arrow-down.svg");
-  background-repeat: no-repeat;
-  background-size: contain;
 }
 
 @media (max-width: 768px) {
@@ -1366,7 +1349,6 @@ h4.tab-no-data {
 .station-timeseries-main h2 + p {
   font-size: 14px;
   line-height: 19px;
-  color: #bdbdbd;
 }
 
 .station-timeseries-additional {
@@ -1450,7 +1432,8 @@ h4.tab-no-data {
 
 .form-list-download {
   display: flex;
-  align-items: center;
+  flex-direction: column;
+  align-items: flex-start;
   flex-wrap: wrap;
   max-width: 560px;
   width: 100%;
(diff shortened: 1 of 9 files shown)


aptiko:
Here are some comments (I know some of them are a design problem, not an implementation problem, it doesn't matter).

1. The breadcrumbs end with a small arrow-down. This gives me the impression that it's a menu and I tried clicking it. I think we can just remove it (I don't understand its use).
2. The dates underneath the variable name are an important information for the user and the user should be able to read them. Therefore they shouldn't be grayed out. Use the default foreground color.
3. If there are no dates (e.g. choose any station except for Agios Spyridonas, no other station has data in the development environment), the dates show badly; you need to remove the line.
4. Remove the Lorem ipsum text.
5. I think the time series remarks would be fine without the "Notes:" heading.
6. Use "Download data" instead of "Download report:", without a colon. For the button, use just "download".
7. You need to add the "Select type" and "Select format" options. However, "Select type" should be "Select time series", and there is a design problem: the design shows four options, but there can actually be unlimited options (there will rarely be more than a few, but there could certainly be five or six). Would you like to propose a solution?

ad0v0:
**Feedback to implement**:

* [ ] The breadcrumbs end with a small arrow-down. This gives me the impression that it's a menu and I tried clicking it. I think we can just remove it (I don't understand its use).
* [ ] The dates underneath the variable name are an important information for the user and the user should be able to read them. Therefore they shouldn't be grayed out. Use the default foreground color.
* [ ] If there are no dates (e.g. choose any station except for Agios Spyridonas, no other station has data in the development environment), the dates show badly; you need to remove the line.
* [ ] Remove the Lorem ipsum text.
* [ ] I think the time series remarks would be fine without the "Notes:" heading.
* [ ] Use "Download data" instead of "Download report:", without a colon. For the button, use just "download".
* [ ] You need to add the "Select type" and "Select format" options. However, "Select type" should be "Select time series", and there is a design problem: the design shows four options, but there can actually be unlimited options (there will rarely be more than a few, but there could certainly be five or six). Would you like to propose a solution?

aptiko:
You have added a file `gallery.html`. This appears to be an error.

In addition, you have made a minor change in `templates/enhydris/station_detail/remarks.html`. Although that minor change is probably correct, it does not belong in this pull request.

aptiko:
Merged in e15989e3a0e85d54a5feea1331e6bc5e2e272d46.

Issue #493, opened 2023-05-10
Internal Server Error in invalid old timeseries URL


aptiko:
https://openmeteo.org/api/stations/1407/timeseries/%7Bstations_temperature_feature_id%5Bstation%5D%7D/data/?fmt=json&exact_datetime=true&start_date=%7Bdatetime_object.date()%7D%20%7Bdatetime_object.time()%7D

invalid literal for int() with base 10: '{stations_temperature_feature_id[station]}'

PR #494, opened 2023-05-19
fixed invalid url old timerseries

Shrhawk:
**Checklist**:

* [ ] All tests are passing

### DIFF
--- a/enhydris/api/views.py
+++ b/enhydris/api/views.py
@@ -167,6 +167,7 @@ class TimeseriesViewSet(ModelViewSet):
     CHART_MAX_INTERVALS = 200
     queryset = models.Timeseries.objects.all()
     serializer_class = serializers.TimeseriesSerializer
+    lookup_value_regex = '\d+'
 
     def get_permissions(self):
         if self.action in ("data", "bottom", "chart"):
@@ -237,7 +238,7 @@ class TimeseriesViewSet(ModelViewSet):
 
     @action(detail=True, methods=["get"])
     def bottom(self, request, pk=None, *, station_id, timeseries_group_id=None):
-        ts = get_object_or_404(models.Timeseries, pk=pk)
+        ts = get_object_or_404(models.Timeseries, pk=int(pk))
         self.check_object_permissions(request, ts)
         response = HttpResponse(content_type="text/plain")
         timezone_param = request.GET.get("timezone", None)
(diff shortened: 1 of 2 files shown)


Shrhawk:
https://github.com/openmeteo/enhydris/issues/493

enhydris/api/views.py
```diff
@@ -237,7 +238,7 @@ def data(self, request, pk=None, *, station_id, timeseries_group_id=None):
 
     @action(detail=True, methods=["get"])
     def bottom(self, request, pk=None, *, station_id, timeseries_group_id=None):
-        ts = get_object_or_404(models.Timeseries, pk=pk)
+        ts = get_object_or_404(models.Timeseries, pk=int(pk))
```
aptiko, line 241: Was this working before? If yes, why is this change needed? (And how on earth was it working?)

enhydris/api/views.py
```diff
@@ -246,7 +247,7 @@ def bottom(self, request, pk=None, *, station_id, timeseries_group_id=None):
 
     @action(detail=True, methods=["get"])
     def chart(self, request, pk=None, *, station_id, timeseries_group_id=None):
-        timeseries = get_object_or_404(models.Timeseries, pk=pk)
+        timeseries = get_object_or_404(models.Timeseries, pk=int(pk))
```
aptiko, line 250: Same question.

enhydris/api/views.py
```diff
@@ -167,6 +167,7 @@ class TimeseriesViewSet(ModelViewSet):
     CHART_MAX_INTERVALS = 200
     queryset = models.Timeseries.objects.all()
     serializer_class = serializers.TimeseriesSerializer
+    lookup_value_regex = '\d+'
```
aptiko, line 170: If this line is added, then I can't reproduce the problem any more. If that is the case, why do we have the `try...except` in `_get_data()`?

aptiko (changes requested):
1. The commit message needs proper capitalization and spelling, and should also be in the imperative. It should also have `(fixes #493)` in the end. (See also "Git and GitHub" [section "Commit and commit messages"] in Confluence.)
2. I found out that this problem also occurs in the new URLs after all, not only in the deprecated; for example, under `/api/stations/2012/timeseriesgroups/584/timeseries/{hello}/data/`. This means the commit message should not mention the "old" URL. It could be, for example, `Fix crash in invalid timeseries URL (fixes #493)`.

enhydris/api/views.py
Shrhawk: Yes @aptiko it was working fine, just make it consistent for the views

enhydris/api/views.py
Shrhawk: @aptiko added try catch at `_get_data()` here bcz it's not the direct part of view, can be called anywhere from the code

Shrhawk:
Closed in favor of https://github.com/openmeteo/enhydris/pull/495 (renamed the branch)

PR #496, opened 2023-05-25
Fix delete station (fixes #492)

### LINKED ISSUE #492: Cannot delete stations
Attempting to delete a station on the admin usually fails, because it tries to warn about every single time series record.

Shrhawk:
**Checklist**:

* [ ] All tests are passing

### DIFF
--- a/enhydris/admin/station.py
+++ b/enhydris/admin/station.py
@@ -331,6 +331,13 @@ class StationAdmin(ObjectPermissionsModelAdmin, nested_admin.NestedModelAdmin):
     search_fields = ("id", "name", "code", "owner__ordering_string")
     list_display = ("name", "owner")
     list_filter = (SiteFilter,)
+    delete_confirmation_template = "admin/enhydris/station/delete_confirmation.html"
+    delete_selected_confirmation_template = (
+        "admin/enhydris/station/delete_selected_confirmation.html"
+    )
+
+    def get_deleted_objects(self, objs, request):
+        return [], {}, set(), []
 
     def get_queryset(self, request):
         if request.user.is_superuser:
--- a/enhydris/templates/admin/enhydris/station/delete_confirmation.html
+++ b/enhydris/templates/admin/enhydris/station/delete_confirmation.html
@@ -1,0 +1,33 @@
+{% extends "admin/delete_confirmation.html" %}
+{% load i18n %}
+{% block content %}
+    {% if perms_lacking %}
+        {% block delete_forbidden %}
+            <p>{% blocktranslate with escaped_object=object %}Deleting the {{ object_name }} '{{ escaped_object }}'
+                would result in deleting related objects, but your account doesn't have permission to delete the
+                following types of objects:{% endblocktranslate %}</p>
+            <ul id="deleted-objects">{{ perms_lacking|unordered_list }}</ul>
+        {% endblock %}
+    {% elif protected %}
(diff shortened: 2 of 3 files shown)


Shrhawk:
Related pull request https://github.com/openmeteo/enhydris-synoptic/pull/81

enhydris/models/timeseries_group.py
```diff
@@ -85,7 +85,7 @@ class Meta:
 
 class TimeseriesGroup(models.Model):
     last_modified = models.DateTimeField(default=now, null=True, editable=False)
-    gentity = models.ForeignKey(Gentity, on_delete=models.CASCADE)
+    gentity = models.ForeignKey(Gentity, on_delete=models.SET_NULL, null=True)
```
aptiko, line 88: This isn't right. All `TimeseriesGroup` objects should have a related `gentity`, it can't be null. I believe that if you search the Django documentation or the web (or if you ask ChatGPT) you'll find a way for admin to not warn you about the deletion of related objects. It is this warning that causes the problem, because it can be huge and it takes a large amount of time to prepare it.

aptiko (changes requested):
Please also properly capitalize the commit message.

enhydris/models/timeseries_group.py
Shrhawk: @aptiko please review now

enhydris/admin/station.py
```diff
@@ -331,6 +331,16 @@ class StationAdmin(ObjectPermissionsModelAdmin, nested_admin.NestedModelAdmin):
     search_fields = ("id", "name", "code", "owner__ordering_string")
     list_display = ("name", "owner")
     list_filter = (SiteFilter,)
+    actions = None
```
aptiko, line 334: Why `actions = None`?

enhydris/admin/station.py
```diff
@@ -331,6 +331,16 @@ class StationAdmin(ObjectPermissionsModelAdmin, nested_admin.NestedModelAdmin):
...
+
+    def delete_view(self, request, object_id, extra_context=None):
+        """
+        Perform the deletion directly without displaying the confirmation page
+        """
+        obj = self.get_object(request, object_id)
+        obj_display = str(obj)
+        self.delete_model(request, obj)
+        return self.response_delete(request, obj_display, object_id)
```
aptiko, line 343: The problem with this is that it will not show a confirmation at all. We do want a confirmation, but we don't want it to show the related objects.

You may want to look at [Django ticket 10919](https://code.djangoproject.com/ticket/10919) to see how other people have worked around this problem.

aptiko (changes requested):

enhydris/admin/station.py
Shrhawk: Skipping the bulk delete here, do we need this @aptiko ?

enhydris/admin/station.py
Shrhawk: Screenshot 2023-06-02 at 3 27 19 PM

@aptiko this looks good to you now ?

aptiko: It's still a problem, because it gives to the user the impression that there are no related objects. So the user might think "OK, this station has nothing in it, no problem if I delete it."

So after the "are you sure?" question, it should say that "All objects related to the station, including images, events and time series, will also be deleted."

Issue #503, opened 2023-07-31
Meteoview2 stores Europe/Athens time instead of converting to UTC


aptiko:

Shrhawk:
When I select UTC in case of Meteoview2, it stores the data_timezone in Etc/GMT which is equal to UTC as per the following [link](https://www.timeanddate.com/worldclock/converter.html?iso=20230926T020000&p1=1440&p2=tz_gmt). Since, there are no differences then why do we need to convert it?

aptiko:
@Shrhawk I'm not certain I understand the question correctly. The Meteoview2 API provides the data in Europe/Athens (actually this is probably configurable for each station, but all the stations we work with are in Greece). But in Enhydris we store data in UTC. So we have to convert from the time zone of the data to UTC. [data_timezone](https://enhydris.readthedocs.io/en/latest/dev/telemetry.html#enhydris.telemetry.models.Telemetry.data_timezone) is the time zone of the data in the API, not the time zone of the data in the database, which is always UTC.

There is also the field [Gentity.display_timezone](https://enhydris.readthedocs.io/en/latest/dev/database.html#enhydris.models.Gentity.display_timezone), but this does not have to do with how the data is stored in Enhydris, only with how it is displayed. The `Telemetry` app does not display any data, it merely downloads it from an API and then stores it in the database, therefore it's not concerned with `display_timezone`.

Now if the Meteoview2 API has an option with which to select in which time zone we want to get the data (like addUPI does), then certainly we must get the data in UTC and store it as is, without conversion, and we don't need to show the `data_timezone` field to the user. (But I'm unaware of any such option.)

Shrhawk:
@aptiko I understand the flow that Meteoview2 API provides the data in Europe/Athens and we have to convert it to UTC and store it to the database. Can you tell me where exactly we are storing the data to the database?

Also, I was checking [get_measurements()](https://github.com/openmeteo/enhydris/blob/master/enhydris/telemetry/types/meteoview2.py#L47), does this method has anything to do with the timezone? If yes, then where are we using it?

aptiko:
@Shrhawk [Here is](https://github.com/openmeteo/enhydris/blob/c6453900701099930d2d8297e1dcd4c00bb5eade/enhydris/telemetry/models.py#L100-L104) where `get_measurements()` is used.

[...]

PR #511, opened 2023-10-08
Fix UTC timezone issue in Meteoview2

Shrhawk:
**Checklist**:

* [ ] All tests are passing

### DIFF
--- a/enhydris/telemetry/tests/test_models.py
+++ b/enhydris/telemetry/tests/test_models.py
@@ -224,7 +224,7 @@ class TelemetryFetchTestCase(TelemetryFetchTestCaseBase):
         )
         data = StringIO()
         timeseries.get_data().write(data)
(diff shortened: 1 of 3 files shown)


Shrhawk:
@aptiko ping, also check the messages

aptiko:
This is not the right way to fix it. The unit tests were actually correct, because the `telemetry.data_timezone` is set by the tests to `Europe/Athens`, meaning it should be converted from that one to UTC, and the `station.display_timezone` is set to `Etc/GMT-2`, which is the same thing for winter, so when retrieved and printed, you get the same result as the value received from the logger.

I think that the main code is also correct and that #503 is invalid and everything works as it's supposed to work, so I'm going to go ahead and close both this and #503. If I'm wrong and there is a problem after all, we'll certainly discover it.

